PR #9, opened 2017-05-18
Add basic Polymer app and pull from fixed results set

jeffcarp:
This PR:
- Adds Polymer via (`bower.json` and vendored `bower_components`, sorry for the messy diff)
- Adds a root component `<wpt-results>` in `components/` and in the HTML in `templates/index.html`

The root component currently loads WPT results from one hardcoded file simply for UI display purposes. I'd like to land this to get Polymer on master and then follow it up with improvements to the loading, searching, and filtering of results.


components/wpt-results.html
```diff
@@ -0,0 +1,107 @@
...
+
+      static get properties() {
+        return {
+          specs: {
+            type: Array,
+            observer: '_specsChanged',
+            value: []
+          }
+        };
+      }
```
qyearsley: These getters are part of the standard way that components are defined in polymer2 I right?

components/wpt-results.html
```diff
@@ -0,0 +1,107 @@
...
+
+      connectedCallback() {
+        super.connectedCallback();
+
+        this._fetchResults()
+      }
+
+      _specFromTestPath(path) {
+        return path.split('/')[1]
+      }
```
qyearsley: This is getting the name of a top-level directory in wpt, right, and the input is expected to start with a slash, right?

Also, are we planning to use a convention of underscore prefixes here to indicate private methods? Is this according to some style guide?

components/wpt-results.html
```diff
@@ -0,0 +1,107 @@
...
+
+      async _fetchResults() {
+        const url = 'https://storage.googleapis.com/wptdashboard.appspot.com/results/e2d2cde03d/chrome-60.0-debian-8.json.gz'
+        const response = await fetch(url)
+        const data = await response.json()
```
qyearsley: I'm new to async/await and some other stuff, so I have some questions:

Do we want to use "await"  for fetch() and response.json() in order to make this not wait and block the main JS thread? (Is there a "main JS thread")?
Do we declare this function "async" in order to use "await" inside of it?

At some point after fetching the results, the value of this.specs changes, and then this._specsChanged is called. Are the HTML elements actually displaying the results in the template above also updated then? If so, is that independent of _specsChanged?

templates/index.html
```diff
@@ -11,7 +11,7 @@
         </nav>
     </header>
     <div>
-      (top-level spec directories will go here)
+      <wpt-results />
```
qyearsley: I think that maybe this should be <wpt-results></wpt-results>

(In html5 I think self-closing tags are not a thing; see http://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5); there are some "void elements" that don't require closing tags, but custom defined elements are never void elements.

components/wpt-results.html
```diff
@@ -0,0 +1,107 @@
...
+<link rel="import" href="/bower_components/polymer/polymer-element.html">
+<link rel="import" href="/bower_components/polymer/lib/elements/dom-repeat.html">
+
+<dom-module id="wpt-results">
+  <template>
+    <style>
+      :host {
+        display: block;
+      }
+      .WPTResults__spec {
```
jeffcarp: It's a naming convention called [BEM](http://getbem.com/naming/). I picked it up when pairing with a designer and it sort of stuck. However, with web components and specifically shadow DOM, BEM's benefits of component reusability pretty much don't matter. Because of this, I'll remove the prefixes from these classes (so it's just `.spec`).

components/wpt-results.html
jeffcarp: Yes sir https://www.polymer-project.org/2.0/docs/devguide/properties

components/wpt-results.html
jeffcarp: I saw underscore methods in a [Polymer 2 codelab](https://codelabs.developers.google.com/codelabs/polymer-2-carousel/index.html?index=..%2F..%2Findex#0) and copied the convention. What do you think?

templates/index.html
jeffcarp: I didn't know they were removed in html5! I'll change that.

components/wpt-results.html
jeffcarp: > Do we want to use "await" for fetch() and response.json() in order to make this not wait and block the main JS thread?

Yep! Both `fetch()` and `response.json()` return promises, so it's mandatory to handle them asynchronously. `await` just accepts a promise, so these two are functionally equivalent:

async/await
```
const response = await fetch(url)
```
promise
```
fetch(url).then(function (response) {
  ... do something with response
})
```

> (Is there a "main JS thread")?

Yes, JS is single threaded and executes in the rendering thread, so if you do expensive DOM operations you can [freeze the renderer](http://jankfree.org/).

> Are the HTML elements actually displaying the results in the template above also updated then? If so, is that independent of _specsChanged?

Yes, the HTML bindings are updated after `this.specs` is changed since that object is being observed (though via what means I'm not sure). So `_specsChanged` fires independently.

Also, this code [won't work out of the box on Edge](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Browser_compatibility), so in a future PR I'll undo the async/await syntax and handle the promises directly.

jeffcarp:
Addressed your PR feedback, this should be ready to go (I'm not sure how GitHub review approvals work or if you need them? the merge button is already green)

qyearsley:
PS: I would guess that by default GitHub reviews don't technically require
approval before merging, but Chromium code review policy
<https://chromium.googlesource.com/chromium/src/+/master/docs/code_reviews.md>
in
general is that all changes should be reviewed.

For cases like this where the reviewer just has minor nits, I think "LGTM
with nits" (approval with comments and suggestions) is good practice and
would avoid blocking you from merging when you're ready.

PR #10, opened 2017-05-26
Add TestRun model and basic Polymer search component

jeffcarp:
Note: some of the Polymer code here isn't meant to be super permanent - just something to get results up and minimally usable on the dashboard. I deployed this branch to [wpt.fyi](http://wpt.fyi) if you want to take a look.

TODO

- [x] Clean up this PR
- [x] Squash branch
- [x] Annotate this PR
- [x] Fix broken browser images
- [x] Remove dom-if


components/wpt-results.html
```diff
@@ -22,20 +22,94 @@
...
+      .browser img {
+        width: 32px;
+        height: 32px;
+      }
     </style>
 
-    <template is="dom-repeat" items="{{specs}}">
+    <section class="search">
+      <input
+        value="{{query::input}}"
```
jeffcarp: This is a Polymer feature for automatically binding the value of the input event to `this.query`

components/wpt-results.html
```diff
@@ -22,20 +22,94 @@
...
+          <div class="browsers">
+            <template is="dom-repeat" items="{{spec.results}}" as="result">
+              <div class="browser">
+                <div class="browser">{{ result.passing }} / {{ result.total }}<br>{{ _passingPercent(result) }}%</div>
+              </div>
+            </template>
+          </div>
+        </div>
+
+        <!-- TODO(jeffcarp): This nested sort isn't working -->
```
jeffcarp: Have been fighting this for a few hours, don't want to waste more time on it.

components/wpt-results.html
```diff
@@ -22,20 +22,94 @@
...
+        <template is="dom-repeat" items="{{spec.testFiles}}" as="testFile" filter="_testFileFilter" sort="_testFileSort" class="test_file_list">
+          <div>{{ testFile.testFile }}</div>
+
+          <!-- This is the only way I've found to get sub-lists to update -->
+          <div style="display:none;">{{query}}</div>
```
jeffcarp: Including `query` in the subtest result sub-tree is the only way I could get the subtest results to update each time `query` is updated. I'll ask someone on the Polymer team if there's a more idiomatic way to do this.

components/wpt-results.html
```diff
@@ -46,59 +120,192 @@
...
       }
 
-      connectedCallback() {
+      async connectedCallback() {
         super.connectedCallback();
 
-        this._fetchResults()
+        const testFileResults = await Promise.all(this.testRuns.map(testRun => {
+          return this._fetchResults(testRun.results_url)
+        }))
```
jeffcarp: tl;dr: `Promise.all` waits for all `_fetchResults` promises to return before continuing on. `testFileResults` is an array of results from each `_fetchResults` call.

components/wpt-results.html
```diff
@@ -46,59 +120,192 @@
...
+        testFileResults.forEach(result => {
+          const testFiles = result.testFiles
+          const resultsURL = result.resultsURL
+
+          Object.keys(result.testFiles).forEach(testFileName => {
+            const specName = this._specFromTestPath(testFileName)
+
+            let spec = this._arrayElemByProperty(specs, 'specName', specName, {
+              results: [], specName: specName, testFiles: []
+            })
```
jeffcarp: This section contains a lot of gymnastics to work with arrays of objects since `dom-repeat` only works with arrays. It would be much easier to simply deal with objects. The next step I want to take is to write my own version of `dom-repeat` that takes an object and supports easier nested repeating to better fit our use case.

components/wpt-results.html
```diff
@@ -46,59 +120,192 @@
...
+        }
+
+        return testFile.testFile.toLowerCase().includes(this.query.toLowerCase())
       }
 
+      _testFileSort(a, b) {
+        if (a.testFile < b.testFile) return -1;
+        if (a.testFile > b.testFil) return 1;
+        return 0;
+      }
```
jeffcarp: These are convenience methods given to `dom-repeat`.

main.go
```diff
@@ -22,7 +22,8 @@ import (
 var templates = template.Must(template.ParseGlob("templates/*.html"))
 
 func init() {
-    http.HandleFunc("/", testHandler)
+    http.HandleFunc("/tasks/populate-dev-data", populateDevData)
```
jeffcarp: `populateDevData` is for boostrapping if you're running the app for the first time in development.

models.go
```diff
@@ -0,0 +1,35 @@
...
+    OSVersion       string          `json:"os_version"`
+
+    // The full SHA1 of the tested WPT revision
+    Revision        string          `json:"revision"`
+
+    // Results URL
+    ResultsURL      string          `json:"results_url"`
+
+    CreatedAt       time.Time       `json:"created_at"`
+}
```
jeffcarp: `TestRun` is a model that isn't (yet) described in the design doc, but I think it fits our use case more optimally. Every time a test run finishes, it'll create a new TestRun. Then, when someone loads a page we can query the latest TestRun by each platform.

test_handler.go
```diff
@@ -15,11 +15,39 @@
...
-    if err := templates.ExecuteTemplate(w, "index.html", nil); err != nil {
-            http.Error(w, err.Error(), http.StatusInternalServerError)
+    ctx := appengine.NewContext(r)
+
+    var testRuns []TestRun
+    var chromeTestRuns []TestRun
+    var firefoxTestRuns []TestRun
+    baseQuery := datastore.NewQuery("TestRun").Order("-CreatedAt").Limit(1)
+    chromeQuery := baseQuery.Filter("BrowserName =", "chrome")
+    firefoxQuery := baseQuery.Filter("BrowserName =", "firefox")
```
jeffcarp: In a later PR I'll parameterize the browser names so it's not hard-coded.

templates/index.html
```diff
@@ -11,7 +12,7 @@
         </nav>
     </header>
     <div>
-      <wpt-results></wpt-results>
+      <wpt-results test-runs="{{ . }}"></wpt-results>
```
jeffcarp: The `[]TestRun` JSON from the handler gets inserted directly into this element's property

models.go
```diff
@@ -0,0 +1,35 @@
...
+    // Platform information
+    BrowserName     string          `json:"browser_name"`
+    BrowserVersion  string          `json:"browser_version"`
+    OSName          string          `json:"os_name"`
+    OSVersion       string          `json:"os_version"`
+
+    // The full SHA1 of the tested WPT revision
+    Revision        string          `json:"revision"`
+
+    // Results URL
```
qyearsley: What is this results URL specifically? Is it a URL of the test results JSON, or a page to view results?

models.go
qyearsley: It seems to make sense. I guess this is how datastore models are generally described in Go?

(In Python, you have to make a class that subclasses ndb.Model...)

Anyway, depending on how we want to query Platform in the future, this model is possible to change later...

templates/index.html
qyearsley: What's the meaning of the dot here btw?

test_handler.go
qyearsley: Ack

main.go
qyearsley: Makes sense.

components/wpt-results.html
qyearsley: Sounds reasonable.

components/wpt-results.html
```diff
@@ -46,59 +120,170 @@
...
-        this._fetchResults()
+        const testFileResults = await Promise.all(this.testRuns.map(testRun => {
+          return this._fetchResults(testRun.results_url)
+        }))
+
+        let specs = []
+
+        testFileResults.forEach(result => {
+          const testFiles = result.testFiles
+          const resultsURL = result.resultsURL
```
qyearsley: Optionally, you could add semicolons on these lines right?

components/wpt-results.html
qyearsley: What does nested sort refer to in this case? Sorting the directories and sorting the tests within the directories?

components/wpt-results.html
qyearsley: Alright! Seems like it will be helpful for me to open a Polymer reference when reading this code :-)

components/wpt-results.html
jeffcarp: Yes, the second - the tests aren't sorted within the spec directories and I'm not sure why

models.go
jeffcarp: Got it. There's actually nothing datastore-specific in this file. It looks like Go App Engine models are just vanilla structs.

templates/index.html
jeffcarp: The dot refers to the top-level object in the template - see the last argument in `templates.ExecuteTemplate`:
https://github.com/GoogleChrome/wptdashboard/pull/10/files#diff-eef986022544a4cde28537615d5c6768R50

For example if the top-level object in the template were a struct, you'd write `{{ .Property }}` to access `struct.Property`.

components/wpt-results.html
jeffcarp: Semicolon use should be a project-wide decision - I used to be a fan of semicolons but was converted after using https://standardjs.com/ - specifically the "No semicolons" bullet point on that page gives a lot of good rationale for leaving them out. I'm open to either style and I know that Google leans more toward semicolon usage.

models.go
jeffcarp: The first one - do you think it should be clarified?

components/wpt-results.html
qyearsley: Ah, interesting! Yep, there's no reason why we have to use semicolons. The most important thing to me is consistency.

(Speaking of which, have you been using eslint and golint for this repo? If you are, feel free to add any relevant config/rc files or notes about their use in a top-level README.md file.)

jeffcarp: I haven't run eslint or golint (or gofmt?) on this repo - I'll make a bug to add those.

Speaking of which - for bugs like adding eslint, are you ok with keeping them in this repo?

qyearsley: You mean using the GitHub issue tracker?

It's probably simpler to keep all issues in one place, so we should decide on either GitHub or Monorail. I'm kind of in favor of Monorail since I'm more used to it :-P

PR #15, opened 2017-06-01
Add pass/fail colors for tests

jeffcarp:
Previously tests had no color coding (green/red). Now they do.

Check this branch out at https://results-colors-dot-wptdashboard.appspot.com

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -98,7 +98,10 @@ limitations under the License.
               <td>{{ testFile.testFile }} <span style="display:none;">{{query}}</span></td>
 
               <template is="dom-repeat" items="{{testFile.results}}" as="result">
-                <td>{{ _passingPercent(result) }}% ({{ result.passing }}/{{ result.total }}) <span style="display:none;">{{query}}</span></td>
+                <td style="{{ _testFileHSL(result) }}">
+                  {{ _passingPercent(result) }}% ({{ result.passing }}/{{ result.total }})
+                  <span style="display:none;">{{query}}</span>
+                </td>
               </template>
             </tr>
           </template>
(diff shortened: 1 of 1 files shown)


components/wpt-results.html
```diff
@@ -284,6 +287,20 @@
       _dateFormat(dateStr) {
         return String(new Date(dateStr)).match(/^\w+ (\w+ \w+ \w+)/)[1]
       }
+
+      _testFileHSL(result) {
```
qyearsley: Optionally, you could also call this _testResultStyle since it returns a CSS style rule string

components/wpt-results.html
```diff
@@ -284,6 +287,20 @@
...
+        // Need saturation between 65-100, reversed (range 35)
+        // hsl(129, [65-100]%, 65%)
```
qyearsley: I'm not sure if this comment is necessary now

components/wpt-results.html
```diff
@@ -284,6 +287,20 @@
...
+        const passRate = result.passing / result.total
+        if (passRate === 1) {
+          // Green
+          return 'background-color: hsl(129, 85%, 65%)'
+        } else {
+          const luminance = 65 + passRate * 20
+          // Some shade of red
+          return `background-color: hsl(0, 85%, ${luminance}%)`
```
qyearsley: Hmm, so hue = 0 is red, and saturation is the same either way, and we're changing from dark red to light red by using a luminance between 65% and 85%?

components/wpt-results.html
jeffcarp: Will do

components/wpt-results.html
jeffcarp: Will remove

components/wpt-results.html
jeffcarp: It should be luminance between 100% (meaning 0% passing) and 65% (meaning 99% passing), it's what @foolip described which led to crbug.com/688863 (which I guess we can close now)

PR #16, opened 2017-06-07
Search refactor

jeffcarp:
This PR:
- Makes search faster by limiting the number of results returned
- Fixes the problem where search wouldn't update for some inputs
- Refactors all data to be stored in objects instead of arrays, then converts the data into an array for display
- Better supports missing results. Right now if you're looking at results from browser 1 and 2, if results for a test file for 1 are missing, 2's results will appear in 1's column.

TODO:

- [x] Sorting is broken in this PR

See this branch at https://search-refactor-dot-wptdashboard.appspot.com

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -15,6 +15,7 @@ limitations under the License.
 -->
 
 <link rel="import" href="/bower_components/polymer/polymer-element.html">
+<link rel="import" href="/bower_components/polymer/lib/elements/dom-if.html">
 <link rel="import" href="/bower_components/polymer/lib/elements/dom-repeat.html">
 
 <dom-module id="wpt-results">
@@ -79,9 +80,7 @@ limitations under the License.
       </thead>
       <tbody>
 
-        <template
-          is="dom-repeat"
-          items="{{specs}}" as="spec" filter="_specFilter" id="spec_list" sort="_specSort">
+        <template is="dom-repeat" items="{{displayedSpecs}}" as="spec" id="spec_list">
 
           <tr class="spec">
             <td><b>{{spec.specName}}</b></td>
@@ -92,7 +91,7 @@ limitations under the License.
           </tr>
 
           <!-- TODO(jeffcarp): This nested sort isn't working -->
-          <template is="dom-repeat" items="{{spec.testFiles}}" as="testFile" filter="_testFileFilter" sort="_testFileSort" class="test_file_list">
+          <template is="dom-repeat" items="{{spec.testFiles}}" as="testFile" class="test_file_list">
             <tr>
               <!-- This is the only way I've found to get sub-lists to update -->
               <td>{{ testFile.testFile }} <span style="display:none;">{{query}}</span></td>
@@ -106,6 +105,9 @@ limitations under the License.
             </tr>
           </template>
 
+          <template is="dom-if" if="{{ spec.notAllTestFilesShown }}">
+            <div>not all matching test files shown</div>
+          </template>
 
         </template>
       </tbody>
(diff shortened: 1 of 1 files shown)


jeffcarp:
Ok this should be ready for review. The latency currently on wpt.fyi is not due to the network but due to the amount of time it takes to build the specs data structure. he performance on wpt.fyi has gone downhill considerably recently - it takes ~30 seconds on my workstation to reach first meaningful paint. This PR speeds things up considerably to 2.2 seconds to FMP.

qyearsley:
Overall looks good but I won't be able to properly review till Monday - might have some questions after reading over this!

jeffcarp:
@qyearsley np!

components/wpt-results.html
```diff
@@ -125,6 +128,10 @@
             observer: '_queryChanged'
           },
           specs: {
```
qyearsley: General naming comment: I think that using the word "specs" isn't quite correct here; top-level directories in wpt are supposed to correspond to specifications, but they are not themselves specs.

"Specs" here does refer to top-level directories in wpt, right? I do appreciate that "specs" is nice and short and generally gives the right idea though. How do you feel about this?

components/wpt-results.html
```diff
@@ -146,40 +153,31 @@
         testFileResults.forEach(result => {
```
qyearsley: What is one "result"? Is it all of the results for one run (on one browser, on one platform)? Does it correspond to one "results.json" file or something like that?

components/wpt-results.html
```diff
@@ -146,40 +153,31 @@
...
           const testFiles = result.testFiles
           const resultsURL = result.resultsURL
+          const encodedResultsURL = btoa(result.resultsURL)
 
           Object.keys(result.testFiles).forEach(testFileName => {
```
qyearsley: What is done for each test file here? It looks like the purpose of this is to update `this.specs` so that it contains the test file and so that the aggregate summary results are updated, is that right?

(More comments might be helpful here, but I'm not sure.)

components/wpt-results.html
```diff
@@ -146,40 +153,31 @@
         testFileResults.forEach(result => {
           const testFiles = result.testFiles
           const resultsURL = result.resultsURL
+          const encodedResultsURL = btoa(result.resultsURL)
```
qyearsley: What is result.resultsURL? Is it the URL for the JSON that contains results for one run? And what's the advantage of using encodedResultsURL over resultsURL below?

components/wpt-results.html
jeffcarp: Yes that's correct - how about `specDirs`?

components/wpt-results.html
jeffcarp: Yep - each `result` in this case is one `results.json`.

components/wpt-results.html
jeffcarp: Yep that's correct, it's just building the hierarchy of specs. I can add more commenting in a followup.

components/wpt-results.html
jeffcarp: There was a reason I was encoding the results URL to base64 that I'm not remembering right now. But yes, that's correct.

qyearsley:
Excellent! So, right, my main thought here is that we want to try to keep this file as readable and clear as possible for new readers who aren't very familiar with the code -- `wpt-results.html` is quite central and important, and inherently complex. Possible improvements:

 * Explain with a comment that `specs` refers to top-level directories in web-platform-tests, which should correspond to specifications. Optionally rename to `specDirs` if you think that name is OK.
 * Explain why `encodedResultsURL` is used
 * Rename result to `singleRunResults` or something like that if it's clearer.
 * Add function comments for long functions like _refreshDisplayedSpecs and connectedCallback.

jeffcarp:
Agreed, these are all good ideas, I'll make a bug for them.

PR #21, opened 2017-06-27
Sort specs and test results

jeffcarp:
This was originally meant to be included in #16 but wasn't uploaded before the PR was landed.

This also deletes two now unused methods.

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -182,7 +182,7 @@ limitations under the License.
 
       _testFileSort(a, b) {
         if (a.testFile < b.testFile) return -1;
-        if (a.testFile > b.testFil) return 1;
+        if (a.testFile > b.testFile) return 1;
         return 0;
       }
 
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -180,23 +180,6 @@ limitations under the License.
         this._refreshDisplayedSpecs()
       }
 
-      _specFilter(spec) {
-        if (this.query.length === 0) {
-          return true
-        }
-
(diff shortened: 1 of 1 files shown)


components/wpt-results.html
```diff
@@ -180,23 +180,6 @@
...
-        }
-
-        const matchesQuery = (tf) => tf.testFile.toLowerCase().includes(this.query.toLowerCase())
-        return spec.testFiles.some(matchesQuery)
-      }
-
-      _testFileFilter(testFile) {
-        if (this.query.length < 3) {
-          return false
-        }
```
qyearsley: This was the thing that was making it so that individual files aren't displayed until the query is at length 3, right? What's the behavior after this CL?

components/wpt-results.html
```diff
@@ -180,23 +180,6 @@
...
-
-        return testFile.testFile.toLowerCase().includes(this.query.toLowerCase())
-      }
-
       _testFileSort(a, b) {
         if (a.testFile < b.testFile) return -1;
         if (a.testFile > b.testFil) return 1;
```
qyearsley: There's a typo here which I think could lead to incorrect sorting behavior if there were a list with some duplicate test file names. This line isn't changed in this PR, but could be fixed here.

components/wpt-results.html
jeffcarp: This was actually unused, the logic also appears in `_refreshDisplayedSpecs`:
https://github.com/GoogleChrome/wptdashboard/blob/master/components/wpt-results.html#L249

So this deletion has no behavioral effect.

components/wpt-results.html
jeffcarp: Great catch! Will fix in this PR.

Issue #20, opened 2017-06-20
Clarify and add comments to WPTResults component


jeffcarp:
Via @qyearsley in #16:

- Explain with a comment that specs refers to top-level directories in web-platform-tests, which should correspond to specifications. Optionally rename to `specDirs` if you think that name is OK.
- Explain why `encodedResultsURL` is used
- Rename result to `singleRunResults` or something like that if it's clearer.
- Add function comments for long functions like `_refreshDisplayedSpecs` and `connectedCallback`.

Issue #11, opened 2017-05-26
Add JavaScript linting


jeffcarp:
This could take the form of eslint or standard JS

PR #32, opened 2017-07-08
Directory filtering

### LINKED ISSUE #27: Have a way of viewing all tests in a directory
Via https://crbug.com/738440

My opinion is that both of the following should be possible:

1. `wpt.fyi/cors` shows all tests in `/cors`
2. `wpt.fyi/?dirs=cors,dom/nodes` shows all tests in `/cors` and `/dom/nodes`

jeffcarp:
This PR fixes #27 and adds:

- The ability to visit a path like `/css/css-align-3`
- Links on all the spec names to these pages

The UI isn't super pretty, but it gets the job done. Any feedback welcome.

Check it out: https://directory-filtering-dot-wptdashboard.appspot.com/css/css-align-3

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -83,7 +83,7 @@ limitations under the License.
         <template is="dom-repeat" items="{{displayedSpecs}}" as="spec" id="spec_list">
 
           <tr class="spec">
-            <td><b>{{spec.specName}}</b></td>
+            <td><b><a href="/{{spec.specName}}">{{spec.specName}}</a></b></td>
 
             <template is="dom-repeat" items="{{spec.results}}" as="result">
               <td>{{ _passingPercent(result) }}% ({{ result.passing }} / {{ result.total }})</td>
@@ -138,11 +138,19 @@ limitations under the License.
           testRuns: {
             type: Array,
           },
+          filteredDirectories: {
+            type: Array,
+            value: []
+          },
         };
       }
 
       async connectedCallback() {
         super.connectedCallback();
+
+        if (window.location.pathname !== '/') {
+          this.filteredDirectories.push(window.location.pathname)
+        }
 
         const testFileResults = await Promise.all(this.testRuns.map(testRun => {
           return this._fetchResults(testRun.results_url)
@@ -229,7 +237,14 @@ limitations under the License.
             return spec.results[key]
           })
 
-          if (this.query.length < 3) {
+          if (this.filteredDirectories.length > 0) {
+              const inFilteredDir = this.filteredDirectories.some(filterDir =>  filterDir.startsWith(`/${displaySpec.specName}`))
+              if (!inFilteredDir) {
+                continue
+              }
+          }
+
+          if (this.query.length < 3 && this.filteredDirectories.length === 0) {
             displaySpec.testFiles = []
             displayedSpecs.push(displaySpec)
             continue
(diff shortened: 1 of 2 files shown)


components/wpt-results.html
```diff
@@ -252,9 +267,18 @@
 
           let matchingTestFiles = allTestFiles.filter(matchesQuery)
           if (matchingTestFiles.length > 0) {
-            displaySpec.testFiles = matchingTestFiles.slice(0, MAX_RESULTS_PER_SPEC)
+
+            if (this.filteredDirectories.length > 0) {
+              displaySpec.testFiles = matchingTestFiles
+              displaySpec.testFiles = displaySpec.testFiles.filter(tf => (
+                this.filteredDirectories.some(filterDir =>  tf.testFile.startsWith(filterDir))
```
jeffcarp: Thanks

components/wpt-results.html
```diff
@@ -229,7 +237,14 @@
             return spec.results[key]
           })
 
-          if (this.query.length < 3) {
+          if (this.filteredDirectories.length > 0) {
+              const inFilteredDir = this.filteredDirectories.some(filterDir =>  filterDir.startsWith(`/${displaySpec.specName}`))
```
jeffcarp: Done

components/wpt-results.html
```diff
@@ -138,12 +138,20 @@
...
             type: Array,
           },
+          filteredDirectories: {
+            type: Array,
+            value: []
+          },
         };
       }
 
       async connectedCallback() {
```
jeffcarp: Yes, it's the callback for when the Polymer element is "connected" to the DOM.

components/wpt-results.html
```diff
@@ -138,12 +138,20 @@
           testRuns: {
             type: Array,
           },
+          filteredDirectories: {
+            type: Array,
+            value: []
```
jeffcarp: Answer for 1&2: yes correct, the reason it's an array is to allow (in the future) spec or API owners (or anyone else) to set up custom views that feature a few different directories, like:

`wpt.fyi/?dirs=css/CSS2/borders,css/CSS2/normal-padding`

components/wpt-results.html
```diff
@@ -138,12 +138,20 @@
...
+          },
         };
```
qyearsley: Not important, but I just noticed a semicolon here and below (after super.connectedCallback()).

Probably not worth bothering about now, but anyway this just reminded me that setting up linting would be neat.

components/wpt-results.html
qyearsley: Got it, and for things like this, readers can look them up in the Polymer documentation, so maybe no need for function comment.

components/wpt-results.html
jeffcarp: Yes definitely, the bug for linting is #11, I hope I can get more time this week to work on WPTD and set up things like that.

components/wpt-results.html
qyearsley: Ah, got it. Optionally could add a short comment here saying that the displayed specs should be the union of the filtered directories (not the intersection).

components/wpt-results.html
jeffcarp: (Also I removed all semicolons from the JS in this PR)

components/wpt-results.html
jeffcarp: I missed this comment before landing this PR but there are some other comments I need to add via #20, so I'll make a separate PR with all of those.

qyearsley:
SGTM

PR #34, opened 2017-07-10
Add Standard JS linting and travis.yml, fix new lint errors

### LINKED ISSUE #11: Add JavaScript linting

jeffcarp:
Resolves #11

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: node_js
+node_js:
+  - "node"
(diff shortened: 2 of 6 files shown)


jeffcarp:
Yay! https://travis-ci.org/GoogleChrome/wptdashboard/builds/252199701

jeffcarp:
Yeah 🤔 the spaces after function names is not super natural to me, but the no semicolons rule seems to have some good reasoning behind it.

qyearsley:
Yep -- the reason why we have style rules and linters is so that we can avoid being distracted by differences in mundane things, and focus on the logic. As long as the style is consistent and readable, it's good.

Issue #47, opened 2017-07-19
Track all current and historical tested browser/OS combinations in one file


jeffcarp:
I think we should have a file in the root that enumerates all browser/OS combinations we currently test (and have ever tested) against. This would help with running the tests by making validating platform & version consistency easier. It would also help anyone who is using our data. One downside is that we'd need to add a new entry for Chrome and Firefox at least once every 6 weeks.

It could be named `browsers.yaml` or `browsers.json` with the following format:

```yaml
browsers:
  - platform_id: chrome-61-linux
    browser_name: chrome
    browser_version: 61.0
    os_name: linux
    os_version: 4.4
  - platform_id: firefox-56-windows-10
    browser_name: firefox
    browser_version: 56.0
    os_name: windows
    os_version: 10
```

Additional components:
- As part of our Python unit tests, validate this file (e.g. all `browser_name` are valid browser names)
- As part of the test runner, require passing in a valid `platform_id`
- Use `browsers.json` to control the initial test run summaries [test_handler.go](test_handler.go) (which should be renamed) loads

PR #50, opened 2017-07-24
Add support for Edge TestRuns

jeffcarp:
After some hacking around I was able to run Edge via Sauce (using the `add-running-script` branch). The full run (even running with `--processes=2`) took 3 days. The main source of latency is that we need to tear down the session every time a test fails - we might want to file a bug to look into optimizing that.

I also cleaned up some of the handlers to be less hard-coded. Once `browsers.json` lands (#47), we should be able to remove any hard-coding from all of the routes.

Specifically this PR:
- Adds the Edge logo to `/static/`
- Adds sample Edge summary JSON data to `/static/`
- Adds an Edge `TestRun` to the entities that get created in `/tasks/populate-dev-data`
- Adds `edge` as a browser that's loaded on all test page loads

I deployed this branch at https://edge-dot-wptdashboard.appspot.com/


populate_dev_data_handler.go
```diff
@@ -23,42 +23,49 @@ import (
     "appengine/datastore"
 )
 
+// Create TestRun entities for local development and testing.
+// These point to JSON files stored in /static/.
+// Key names for each entity are specified to make this route idempotent.
 func populateDevData(w http.ResponseWriter, r *http.Request) {
     ctx := appengine.NewContext(r)
```
qyearsley: Question, unrelated to this PR: Do you know what purpose these Context structs have? Like, would you ever want to make more than one "Context" for a Request?

populate_dev_data_handler.go
```diff
@@ -23,42 +23,49 @@ import (
...
-    // Add Chrome TestRun
-    testRun := &TestRun{
-        Revision: "e2d2cde03d",
-        BrowserName: "chrome",
-        BrowserVersion: "60.0",
-        OSName: "debian",
-        OSVersion: "8",
-        ResultsURL: "/static/chrome-60.0-debian-8.json",
-        CreatedAt: time.Now(),
+    devData := map[string]*TestRun{
```
qyearsley: Noob Go question:

How should I read `map[string]*TestRun`? I guess it's a map of strings to TestRun pointers? It seems somehow strange that *TestRun is after the brackets, since in other languages like C++ this would be more like `map<string, *TestRun>`.

test_handler.go
```diff
@@ -26,20 +26,19 @@ func testHandler(w http.ResponseWriter, r *http.Request) {
...
+    browserNames := []string{"chrome", "edge", "firefox"}
+
+    for _, browserName := range browserNames {
+        var testRunResults []TestRun
+        query := baseQuery.Filter("BrowserName =", browserName)
+        if _, err := query.GetAll(ctx, &testRunResults); err != nil {
+            http.Error(w, err.Error(), http.StatusInternalServerError)
+            return
+        }
+        testRuns = append(testRuns, testRunResults...)
```
qyearsley: Another Go question: is the `myfunc(myslice...)` syntax similar to the Python `myfun(*mylist)` syntax, where the items inside the slice are all put in the argument list for the function invocation?

test_handler.go
```diff
@@ -26,20 +26,19 @@ func testHandler(w http.ResponseWriter, r *http.Request) {
...
     baseQuery := datastore.NewQuery("TestRun").Order("-CreatedAt").Limit(1)
-    chromeQuery := baseQuery.Filter("BrowserName =", "chrome")
-    firefoxQuery := baseQuery.Filter("BrowserName =", "firefox")
-
-    if _, err := chromeQuery.GetAll(ctx, &chromeTestRuns); err != nil {
-        http.Error(w, err.Error(), http.StatusInternalServerError)
-    }
-    if _, err := firefoxQuery.GetAll(ctx, &firefoxTestRuns); err != nil {
-        http.Error(w, err.Error(), http.StatusInternalServerError)
+    // TODO(jeffcarp): get these from browsers.json once it lands on master
```
qyearsley: Is browsers.json going to be in Google Storage?

jgraham:
> The main source of latency is that we need to tear down the session every time a test fails - we might want to file a bug to look into optimizing that.

`--no-restart-on-unexpected`

populate_dev_data_handler.go
jeffcarp: I think contexts are used one-per-request in Go because the language doesn't support traditional OOP.

I found this explanation which has slightly more context (pun very much intended:1st_place_medal: :boom:)

> At Google, we developed a context package that makes it easy to pass request-scoped values, cancelation signals, and deadlines across API boundaries to all the goroutines involved in handling a request.

https://blog.golang.org/context

populate_dev_data_handler.go
jeffcarp: Right that's correct, however as to why it's that way, I'm probably not the best person to ask.

test_handler.go
jeffcarp: It'll be checked into the repo - see #47 for more details

test_handler.go
jeffcarp: Yep exactly, example here: https://gobyexample.com/variadic-functions

jeffcarp:
@jgraham I'll try that out! Do you know if that noticeably affects test result outcomes?

PR #51, opened 2017-07-25
Add browser list, validation tests, and use it in TestHandler

### LINKED ISSUE #47: Track all current and historical tested browser/OS combinations in one file

jeffcarp:
Resolves #47

I decided to split this change out of #17 since it's ready to land and useful.

This PR:
- Adds `browsers.json`, a file listing all platforms we intend to test and display
- Splits Travis into Node and Python builds
- Adds a Python unit test that validates the values of `browsers.json`
- Changes the main `testHandler` to read `browsers.json` and display only those marked as  "initially_loaded"

### DIFF
--- a/run/browsers_test.py
+++ b/run/browsers_test.py
@@ -1,7 +1,21 @@
-import unittest
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import json
 import os
 import re
+import unittest
 
 
 VALID_PLATFORM_ID_REGEX = r"^[a-z0-9\-\.]+$"
@@ -16,11 +30,11 @@ REQUIRED_PLATFORM_FIELDS = {
 
 class TestBrowsers(unittest.TestCase):
 
-    def __init__(self, *args, **kwargs):
-        super(TestBrowsers, self).__init__(*args, **kwargs)
+    @classmethod
+    def setUpClass(cls):
         filepath = os.path.join(os.path.dirname(__file__), os.pardir, 'browsers.json')
         with open(filepath) as f:
-            self.browsers = json.load(f)
+            cls.browsers = json.load(f)
 
     def test_all_platforms_have_only_valid_characters(self):
         for platform_id, platform_info in self.browsers.iteritems():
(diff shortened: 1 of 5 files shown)


run/browsers_test.py
```diff
@@ -0,0 +1,52 @@
...
+            self.assertTrue(re.match(VALID_PLATFORM_ID_REGEX, platform_id),
+                'platform_id with invalid characters: %s' % platform_id)
+
+    def test_all_browsers_have_required_fields(self):
+        for platform_id, platform_info in self.browsers.iteritems():
+            for key, valid_values in REQUIRED_PLATFORM_FIELDS.iteritems():
+                self.assertTrue(key in platform_info.keys(),
+                    'Required field missing: %s (platform %s)' % (key, platform_id))
+
+                if REQUIRED_PLATFORM_FIELDS[key] != None:
```
qyearsley: Nit: Use `is not None` rather than `!= None`

run/browsers_test.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+import json
+import os
+import re
```
qyearsley: We could sort the imports; also, do we want to add a license header here?

run/browsers_test.py
```diff
@@ -0,0 +1,52 @@
...
+    'os_version': None,
+}
+
+class TestBrowsers(unittest.TestCase):
+
+    def __init__(self, *args, **kwargs):
+        super(TestBrowsers, self).__init__(*args, **kwargs)
+        filepath = os.path.join(os.path.dirname(__file__), os.pardir, 'browsers.json')
+        with open(filepath) as f:
+            self.browsers = json.load(f)
```
qyearsley: Could this be put in setUpClass instead of __init__?

run/browsers_test.py
jeffcarp: Yes, good catch on both counts

run/browsers_test.py
jeffcarp: Done

run/browsers_test.py
jeffcarp: Ah that works, some (probably unreliable) source led be to believe that functionality was a Python 3+ thing

PR #52, opened 2017-07-25
Fix Google Analytics page tracking

jeffcarp:
Now that all results pages act like a SPA, page view events are
no longer being sent to Google Analytics. This fixes that by
manually sending the event.

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -322,6 +322,11 @@ limitations under the License.
         this.query = ''
         this._refreshDisplayedSpecDirs()
         window.history.pushState({}, '', path)
+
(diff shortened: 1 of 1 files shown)


components/wpt-results.html
```diff
@@ -322,6 +322,9 @@
         this.query = ''
         this._refreshDisplayedSpecDirs()
         window.history.pushState({}, '', path)
+        if ('ga' in window) {
```
qyearsley: It might help some future readers to add a little line comment here to say that `ga` is Google Analytics.

jeffcarp: Will do.

PR #61, opened 2017-07-27
Stop displaying passing percentages (fixes #59)

### LINKED ISSUE #59: Stop displaying percentages of tests passing
The current display of percentages is misleading in many cases (e.g. webgl where there are fewer tests being run in some implementations than others because of some preconditions failing, or any case where the tests changed between the testsuite revisions used by different browsers). It also makes it more tempting for people to make naive comparisons, of the kind that we are explicilty trying to avoid, lest wpt submissions become subject to marketing-imposed restrictions. I think we should remove all the percentages, and maybe highlight where the number of tests differs between implementations.

jeffcarp:
They are not super useful for a couple reasons:
- The number of tests run between platforms can differ
- Numbers of subtests between tests differ, making comparisons hard

Additionally they pose a risk of being picked up by marketing teams.

This was suggested by @jgraham in #59 and I agree and think it's a good idea. @RByers @drufball @qyearsley would like to get your thoughts on this.

This branch is deployed at https://no-percentages-dot-wptdashboard.appspot.com/

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -107,7 +107,7 @@ limitations under the License.
               <td><b><a href="/{{specDir.specName}}" on-click="_navigate">{{specDir.specName}}</a></b></td>
 
               <template is="dom-repeat" items="{{specDir.results}}" as="result">
-                <td>{{ _passingPercent(result) }}% ({{ result.passing }} / {{ result.total }})</td>
+                <td>{{ result.passing }} / {{ result.total }}</td>
               </template>
             </tr>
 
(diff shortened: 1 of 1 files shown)


RByers:
Yep, LGTM.  I don't see how we could avoid having the absolute totals, but this looks like a good compromise to me.

jgraham:
Well there are ways to avoid the absolute totals e.g. only showing tests that fail in at least one browser by default. I think that's less important, but it would be interesting to work out how to focus the default UI on the information that's most useful for a browser developer looking to fix interop issues.

jeffcarp:
Merged as 614c2628a9d1dcd478efe54096e16c6f860e429a

PR #62, opened 2017-07-27
Add Safari support

jeffcarp:
Now that we have one Sauce run in, we can add Safari results.

Deployed at: https://safari-dot-wptdashboard.appspot.com/


jeffcarp:
Merged as 34cc78df13aee889fe442d2fa024f3578b38acee

foolip:
I'm curious, how long does a full run over Sauce take?

jeffcarp:
@foolip it takes about 3 days as-is. However as @jgraham suggested in https://github.com/GoogleChrome/wptdashboard/pull/50#issuecomment-317548597, using `--no-restart-on-unexpected` speeds things up considerably at the risk of potentially introducing more flakiness.

jeffcarp:
Update: using `--no-restart-on-unexpected` cuts the time down to around 3 hours. Now that I have a run with and without, I'll compare the two.

foolip:
Wow, that's quite a massive difference. Does it mean that the browser will restart fully after any test that wasn't entirely passing? How many restarts are we talking about?

jgraham:
Yeah, it means we restart the browser after all tests that don't pass. The default behaviour is really intended for situations where you have expectations for all tests, and only restart when something unexpected happens because means you're in some unknown state. But having no expectation files is equivalent to expecting all tests to fully pass, so it doesn't work well for the dashboard usecase. I suggest that using `--no-restart-on-unexpected` is a good idea even if it's a little more flaky (I'm not sure if it will be), unless you go ahead and generate expectation files from the previous runs to cut down on restarts.

foolip:
A comparison of the results would be really helpful, looking forward to that. I guess it's possible in principle for a test to put the browser in a bad state causing all following tests to time out, and other unfun things, but if it's not actually happening a 24-fold savings is pretty compelling :)

jeffcarp:
Unfortunately it looks like tests either weren't run or weren't uploaded for the last 1/3 of the results for both Safari and Edge: http://wpt.fyi/ (both of which were using `--no-repeat-on-unexpected`). Once I get a full run I'll do the comparison.

jgraham:
I should also mention that we use `--run-by-dir=3` in our CI to limit the effects of badly behaved tests. This restarts the browser per directory up to three levels deep, so that tests can't have an unexpected influence on so many other tests. I think for wpt it doesn't make much of a difference (less tahn e.g. the browser UI tests where it was a huge win), but it might be worth enabling here anyway.

jeffcarp:
For the most recent Edge run I tried `--run-by-dir=3` and that seems to have fixed the problem, thanks @jgraham! Trying a run on Safari with the same arguments now.

PR #66, opened 2017-08-01
Fix test file results URL building

### LINKED ISSUE #65: Chrome and Firefox are missing subtest results
See below (or any other test result page):
https://wptdashboard.appspot.com/2dcontext/building-paths/canvas_complexshapes_arcto_001.htm

jeffcarp:
Fixes #65 

Since I removed the requirement that platform IDs have a regular structure in #47, test file results now don't show up on individual test pages because they were relying on that structure. 

This is a quick fix. We should have a better solution long-term, like storing the platform ID inside the TestRun.

Compare:
- https://wptdashboard.appspot.com/domparsing/DOMParser-parseFromString-html.html
- https://fix-test-file-results-url-dot-wptdashboard.appspot.com/domparsing/DOMParser-parseFromString-html.html


jeffcarp:
Done

PR #69, opened 2017-08-03
Only show immediate children in directories

### LINKED ISSUE #57: When viewing a directory, only show test files and directories in that directory
Right now if you view `/css`, it shows _all_ files under `/css`. Viewing that page should only show the names of the directories immediately under `/css`, like `/css-transitions-1`.

### LINKED ISSUE #60: Memory usage of long result lists
I tried opening wpt.fyi and browsing css directory on my old personal laptop yesterday and my system froze. It seems that page needs more than 1GB memory on Chrome (as shown in Chrome task manager).

jeffcarp:
Previously when viewing a subdirectory (e.g. /css), the UI would show all test files in all subdirectories of that directory. This PR only shows immediate children of directories, so for instance, `/css` now only shows immediate subdirectories like `/css/css-grid-1`.

Preview up here:
https://limit-directories-dot-wptdashboard.appspot.com/

Fixes #57 
Fixes #60

TODO:

- [x] Get search back and working
- [x] Don't colorize tests on front page (it's basically a sea of red)
- [x] Going to `/css` still shows entries for `/css-paint-api/geometry-background-image-003.html`
- [x] Startup time is a little slower, can it be sped up? (it's actually faster, see below)

Performance impact on time to first usable render: -0.05%
- https://wptdashboard.appspot.com/
- https://limit-directories-dot-wptdashboard.appspot.com/ - 1.85 seconds

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -222,21 +222,21 @@ limitations under the License.
       }
 
       _refreshDisplayedNodes () {
-        /* Recomputes the list of displayed spec directories and their child test results. */
-        const displayedNodeDict = {}
+        /* Recomputes the list of displayed directories and test files. */
+        const displayedNodeMap = new Map()
         const currentPathParts = this._splitIntoLinkedParts(this.path)
 
         const updateResults = (testFileName, dirPath, isDir) => {
-          if (!(dirPath in displayedNodeDict)) {
-            displayedNodeDict[dirPath] = {isDir: isDir, results: {}}
+          if (!displayedNodeMap.has(dirPath)) {
+            displayedNodeMap.set(dirPath, {isDir: isDir, results: {}})
           }
           let results = this.testFiles[testFileName]
           Object.keys(results).forEach(resultURL => {
-            if (!(resultURL in displayedNodeDict[dirPath].results)) {
-              displayedNodeDict[dirPath].results[resultURL] = {passing: 0, total: 0}
+            if (!(resultURL in displayedNodeMap.get(dirPath).results)) {
+              displayedNodeMap.get(dirPath).results[resultURL] = {passing: 0, total: 0}
             }
-            displayedNodeDict[dirPath].results[resultURL].passing += results[resultURL][0]
-            displayedNodeDict[dirPath].results[resultURL].total += results[resultURL][1]
+            displayedNodeMap.get(dirPath).results[resultURL].passing += results[resultURL][0]
+            displayedNodeMap.get(dirPath).results[resultURL].total += results[resultURL][1]
           })
         }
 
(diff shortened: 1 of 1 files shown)


components/wpt-results.html
```diff
@@ -248,64 +218,70 @@
       }
 
       _queryChanged () {
-        this._refreshDisplayedSpecDirs()
+        this._refreshDisplayedNodes()
       }
 
-      _refreshDisplayedSpecDirs () {
+      _refreshDisplayedNodes () {
         /* Recomputes the list of displayed spec directories and their child test results. */
```
qyearsley: In this method comment: "spec directories" -> "nodes"

In general, is a node always a directory or can it also be a full test path?

components/wpt-results.html
```diff
@@ -248,64 +218,70 @@
...
+            }
+          }
 
-            displayTestFile.testFile = originalTestFile.testFile
-            displayTestFile.results = this.testRuns.map(testRun => originalTestFile.results[testRun.results_url])
+          let parts = this._splitIntoLinkedParts(testFileName)
 
-            return displayTestFile
-          })
+          // Catch any top-level directories
```
qyearsley: By "catch", we mean add them to the displayedNodeDict, right?

components/wpt-results.html
```diff
@@ -248,64 +218,70 @@
...
-        const matchesQuery = (tf) => tf.testFile.toLowerCase().includes(this.query.toLowerCase())
-        const displayedSpecDirs = []
-
-        for (var key in this.specDirs) {
-          let specDir = this.specDirs[key]
-          let displaySpecDir = {}
-
-          displaySpecDir.specName = specDir.specName
-          displaySpecDir.results = this.testRuns.map(testRun => specDir.results[testRun.results_url])
+        const displayedNodeDict = {}
```
qyearsley: Bikeshedding: "Dict" is kind of a Python-y term.

This could also be made a Map and called `displayedNodeMap`. There wouldn't be a huge advantage here, but then instead of `Object.keys(displayedNodeDict)` you'd have `displayedNodeDict.keys()` and instead of `!(dirPath in displayedNodeDict)` you'd have `!displayedNodeMap.has(dirPath)`.

components/wpt-results.html
jeffcarp: Done. It can be either - directories and paths are displayed intermixed with each other now. Rephrased to `Recomputes the list of displayed directories and test files.`.

components/wpt-results.html
jeffcarp: Rephrased comments

components/wpt-results.html
jeffcarp: Great suggestion, I simplified it to `displayedNodes` and changed it to a `Map`.

PR #70, opened 2017-08-03
Document in README how to access data

### LINKED ISSUE #55: Document how to access the publicly available JSON data
We store two types of JSON data files in Google Storage:

1. **Test run summary files**

This is one large object where every test file in a run is the key and its value is a list `[passing, total]` of its subtests.

Example: https://storage.googleapis.com/wptdashboard.appspot.com/results/0d78ba20b4/firefox-56.0-linux-3.16.0-4-amd64.json.gz

2. **Individual test file results**

This is an object describing the test status of one file for a run and the status of all its subtests.

Example: https://storage.googleapis.com/wptd/dd44fd07c5/edge-15-windows-10/bluetooth/idl-Bluetooth.html

All of this data is publicly accessible but is undocumented. We should add a documentation section to the site or in the repo describing how to access these results.

jeffcarp:
Resolves #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Structure:
 
 There is no public API for TestRuns, so if you need to access only the most recent results, looking at
 the main page will give you the latest test SHAs. If you need to access earlier results, an
-exhaustive search is the only way to do that (see issue #73 and #43).
+exhaustive search is the only way to do that (see issue [#73](https://github.com/GoogleChrome/wptdashboard/issues/73) and [#43](https://github.com/GoogleChrome/wptdashboard/issues/43)).
 
 ## Miscellaneous
 
--- a/README.md
+++ b/README.md
@@ -20,17 +20,16 @@ See [CONTRIBUTING.md](/CONTRIBUTING.md) for more information on local developmen
 
 ## Using the data
 
-All test result data is public! There are two types of gzipped JSON data files we store:
+All test result data is public. There are two types of gzipped JSON data files we store.
 
-**Test run summary files**
+### Test run summary files
 
 These are of the pattern: `{sha[0:10]}/{platform_id}-summary.json.gz`
 
 - `sha[0:10]` - the first 10 characters of the WPT SHA that run was tested against
 - `platform_id` - the key of the platform configuration in `browsers.json`
 
-Example:
-https://storage.googleapis.com/wptd/791e95323d/firefox-56.0-linux-summary.json.gz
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,62 @@ dev_appserver .
 
 See [CONTRIBUTING.md](/CONTRIBUTING.md) for more information on local development.
 
+## Using the data
+
+All test result data is public. There are two types of gzipped JSON data files we store.
```
qyearsley: -> There are two types of gzipped JSON data files we store: test run summary files, and individual test result files.

README.md
```diff
@@ -18,6 +18,62 @@ dev_appserver .
...
+
+### Test run summary files
+
+These are of the pattern: `{sha[0:10]}/{platform_id}-summary.json.gz`
+
+- `sha[0:10]` - the first 10 characters of the WPT SHA that run was tested against
```
qyearsley: I'm not sure if it's clearer, but I'd write "WPT commit hash" instead of "WPT SHA". And perhaps we could use a colon instead of a hyphen/minus (here and below).

README.md
```diff
@@ -18,6 +18,62 @@ dev_appserver .
...
+- `platform_id` - the key of the platform configuration in `browsers.json`
+
+Example: https://storage.googleapis.com/wptd/791e95323d/firefox-56.0-linux-summary.json.gz
```
qyearsley: Examples are always helpful :-)

If it's important and expected to say the same, we could also note that the bucket is `wptd`. (Is that the bucket name?)

README.md
```diff
@@ -18,6 +18,62 @@ dev_appserver .
...
+        }
+    ]
+}
+```
+
+### Large-scale analysis
+
+There is currently no public API for TestRuns, so if you need to access only the most recent results, looking at
+the main page will give you the latest test SHAs. If you need to access earlier results, currently an
+exhaustive search is the only way to do that.
```
qyearsley: Statements with "currently" are at a higher risk of getting out-of-date. If there's a chance somebody will add a public API for TestRuns in the future, you could file an issue and point to it here.

README.md
jeffcarp: Done

README.md
jeffcarp: Done x2

README.md
jeffcarp: Yes and done.

README.md
jeffcarp: Done

PR #71, opened 2017-08-03
Add Python linting

### LINKED ISSUE #67: Add Python linting
Via https://github.com/GoogleChrome/wptdashboard/pull/17/files#r129649664

jeffcarp:
Resolves #67

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 requests==2.18.1
+pep8==1.7.0
 
 # This is needed to upload results files
 # google-cloud==0.26.1
--- a/run/browsers_test.py
+++ b/run/browsers_test.py
@@ -51,8 +51,7 @@ class TestBrowsers(unittest.TestCase):
                 self.assertTrue(
                     key in platform_info.keys(),
                     'Required field missing: %s (platform %s)' %
-                    (key, platform_id)
-                )
+                    (key, platform_id))
 
                 if REQUIRED_PLATFORM_FIELDS[key] is not None:
                     self.assertTrue(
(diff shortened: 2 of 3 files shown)


run/browsers_test.py
```diff
@@ -28,28 +28,37 @@
...
+                    (key, platform_id)
+                )
 
                 if REQUIRED_PLATFORM_FIELDS[key] is not None:
-                    self.assertTrue(platform_info[key] in REQUIRED_PLATFORM_FIELDS[key],
-                        'Field has invalid value: %s (platform %s)' % (key, platform_id))
+                    self.assertTrue(
+                        platform_info[key] in REQUIRED_PLATFORM_FIELDS[key],
+                        'Field has invalid value: %s (platform %s)'
+                        % (key, platform_id))
```
qyearsley: PEP8 accepts the closing function call parenthesis either on its own line or on the previous line, as we can see on line 61 and 55.

Do you have any opinions about whether you like one way or the other better?

(Not really important for this PR, just thought I'd ask)

jeffcarp: I don't have an opinion (which might be why I was inconsistent in this PR).

I'll go with keeping the closing parenthesis on the previous line since that looks a little more pythonic. I'll update the PR.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 requests==2.18.1
+pep8==1.7.0
```
gsnedders: `pep8` has been renamed to `pycodestyle` (but not aliased, le sigh, and no warning or anything…). @jeffcarp can you write a PR to change this to latest pycodestyle?

jeffcarp: k

Issue #73, opened 2017-08-04
Give people a way to fetch an index of all TestRuns


jeffcarp:
I'm about to land #70 describing how to access public data if you know the parameters beforehand, but there's no way to retrieve historical data without doing an exhaustive search of all possible SHAs and platform IDs.

This could take the form of a scheduled job that enumerates all uploaded data and uploads a file `testruns.json` to the root of the bucket.

jeffcarp:
An index is now available at https://storage.googleapis.com/wptd/testruns-index.json but that file is not regularly updated.

JKereliuk:
How often are the tests run? Is there any way to currently access what SHA/platforms are available with date information (other than looking up the date from the first 10 characters of the SHA).

lukebjerring:
This issue was moved to web-platform-tests/wpt.fyi#30

Issue #56, opened 2017-07-25
Discuss moving to a static site


jeffcarp:
This is possible and would greatly improve WPT Dashboard's portability. This isn't super high priority right now but I'm creating a bug to better map out what the work would look like.

The work as I understand it would be:
- Transition from storing `TestRuns` in AE Datastore to generating them with a cron job and storing them in a file in GS for the front-end to load
- Re-write `about.html` to be a Polymer component
- Set up static serving - it'd be fantastic if we could serve the site via `gh-pages`
  - This would eliminate the need to set up a separate deployment step (#39)

jeffcarp:
Closing for now

PR #74, opened 2017-08-04
Add script for generating TestRun index

jeffcarp:
This solves half of #73. The other half is regularly running it. (However I don't have plans to set that up yet, I think we should see what the demand is.)

As a side note, if we set up this script to run regularly, we then would have no need to store `TestRun`s in Datastore and could transition the site to be completely static (issue #56).

### DIFF
--- a/util/generate_testrun_index.py
+++ b/util/generate_testrun_index.py
@@ -18,10 +18,10 @@ import json
 from google.cloud import storage
 
 """
-Scans all WPT results directories and generates an index.
+Scans all WPT results directories then generates and uploads an index.
 
-You must be logged into gcloud to run this script. The non-public permission
-necessary to generate this index is `bucket.list_blobs`.
+You must be logged into gcloud and a member of the wptdashboard project
+in order for this script to work.
 """
 
 GCP_PROJECT = 'wptdashboard'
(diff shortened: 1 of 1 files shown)


util/generate_testrun_index.py
```diff
@@ -0,0 +1,72 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+from google.cloud import storage
+
+"""
+Scans all WPT results directories and generates an index.
+
+You must be logged into gcloud to run this script. The non-public permission
```
qyearsley: 1. Is this permission something that the wptdashboard project owner has to grant?
2. Would it also sound OK to say "The permission..." instead of "The non-public permission..."?

util/generate_testrun_index.py
```diff
@@ -0,0 +1,72 @@
...
+    by_sha = {}
+    by_platform = {}
+
+    iterator = bucket.list_blobs(delimiter="/")
+    response = iterator._get_next_page_response()
+    sha_directories = response['prefixes']
+
+    for sha_directory in sha_directories:
+        sha = sha_directory.replace('/', '')
+        iterator = bucket.list_blobs(delimiter="/", prefix=sha_directory)
```
qyearsley: Nit: Could use single quotes or double quotes everywhere for consistency.

util/generate_testrun_index.py
```diff
@@ -0,0 +1,72 @@
...
+        response = iterator._get_next_page_response()
+        platform_directories = [
+            prefix[len(sha_directory):].replace('/', '')
+            for prefix in response['prefixes']]
```
qyearsley: Nit: Here I'd argue the more common Python style is to put the closing `]` on its own line (like the `}` of the dict below).

util/generate_testrun_index.py
```diff
@@ -0,0 +1,72 @@
...
+
+GCP_PROJECT = 'wptdashboard'
+RESULTS_BUCKET = 'wptd'
+
+
+def main():
+    storage_client = storage.Client(project=GCP_PROJECT)
+    bucket = storage_client.get_bucket(RESULTS_BUCKET)
+    by_sha = {}
+    by_platform = {}
```
qyearsley: For `by_sha` and `by_platform`, you could have a comment above explaining what these are for.

It looks like `by_sha` maps hashes (of what I'm not sure - wpt commits?) to lists of platforms; and `by_platform` maps platform strings to lists of hashes. These two things are serialized as JSON and uploaded as "testruns-index.json" (which will be overwritten/updated when this script is rerun?)

util/generate_testrun_index.py
jeffcarp: Rephrased to "You must be logged into gcloud and a member of the wptdashboard project in order for this script to work."

util/generate_testrun_index.py
jeffcarp: Done

util/generate_testrun_index.py
jeffcarp: Done

util/generate_testrun_index.py
jeffcarp: Yes - added documentation.

PR #76, opened 2017-08-09
WIP - Add script for calculating interoperable subset of the web

jeffcarp:
TODO

- [ ] Fix lint
- [ ] Don't hard-code TestRun indexes (this will require adding date to indexes)


jeffcarp:
Closing this for now. I think a starter for data exploration should be provided in the form of a Python notebook. This would be a good candidate for some starter code.

PR #77, opened 2017-08-09
Sauce support

jeffcarp:
This PR contains a really hacky workaround for passing the Sauce platform to `wptrunner`. The command line argument `--sauce-platform` doesn't allow spaces, so to work around that we directly patch WPT to hard-code the value. Bug: https://github.com/w3c/web-platform-tests/issues/6852

The long term resolution for the aforementioned bug would be either:
- Fix it upstream (potentially by splitting up the OS name and OS version as arguments and concatenating them in WPT)
- Move to calling `wptrunner` in Python instead of via shell


run/run.py
```diff
@@ -345,6 +358,28 @@ def get_config():
     return config['default']
 
 
+def patch_wpt(config, platform):
+    '''Applies util/wpt.patch to WPT.
+
+    The patch is necessary to keep WPT running on long runs.
+    jeffcarp has a PR out with this patch:
+    https://github.com/w3c/web-platform-tests/pull/5774
+    '''
```
qyearsley: Nit: double-quotes are more common for docstrings, but it's not really important.

run/run.py
```diff
@@ -345,6 +358,28 @@ def get_config():
...
+    patch_path = '%s/util/wpt.patch' % config['wptd_path']
+    with open(patch_path) as f:
+        patch = f.read()
+
+    # Hackery since the --sauce-platform command line arg doesn't
```
qyearsley: Nit: You could perhaps remove "Hackery since". The fact that you have to explain it in a comment implies that there's some hackiness :-P

run/running.example.ini
```diff
@@ -12,3 +12,7 @@ wptd_prod_host = https://running-dot-wptdashboard.appspot.com
 gs_results_bucket = wptd
 install_wptrunner = false
 secret = SECRET_UPLOAD_TOKEN_GOES_HERE_YOU_NEED_THIS_TO_CREATE_TEST_RESULTS
+sauce_connect_path = TODO
+sauce_tunnel_id = TODO
+sauce_user = TODO
+sauce_key = TODO
```
qyearsley: Rather than TODO, you could use something like TO_BE_FILLED_IN

run/run.py
```diff
@@ -345,6 +358,28 @@ def get_config():
...
+    # accept spaces, but Sauce requires them in the platform name.
+    patch = patch.replace('__platform_hack__', '%s %s' % (
+        platform['os_name'], platform['os_version'])
+    )
+
+    cmd = ['git', 'apply', '-']
```
qyearsley: This variable is probably not necessary (can be inlined), and is also inconsistent since above you use a variable name `command`.

run/run.py
jeffcarp: Done

run/run.py
jeffcarp: Done

run/run.py
jeffcarp: Done

run/running.example.ini
jeffcarp: Done

PR #82, opened 2017-08-16
Add --run-by-dir and --no-restart-on-unexpected to Sauce

jeffcarp:
As mentioned in https://github.com/GoogleChrome/wptdashboard/pull/62#issuecomment-322883386, adding `--run-by-dir` fixes the issue where the run stops producing results 2/3 of the way through.


jeffcarp:
Going to land this so that instance-2 runs it at midnight tonight.

PR #102, opened 2017-08-26
Increase row height for easier clicking on mobile devices

TimvdLippe:
On mobile devices, clicking on the individual tests/categories was quite hard. Increase the row height to 30px makes it a lot easier to click on it.

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -64,6 +64,12 @@ limitations under the License.
         padding: 0.2em 0.5em;
         border: solid 1px #ccc;
       }
+      table tr td:first-child::after {
+        content: "";
+        display: inline-block;
+        vertical-align: top;
+        min-height: 30px;
+      }
       .path-separator {
         padding: 0 0.1em;
       }
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -64,12 +64,6 @@ limitations under the License.
         padding: 0.2em 0.5em;
         border: solid 1px #ccc;
       }
-      table tr td:first-child::after {
-        content: "";
-        display: inline-block;
-        vertical-align: top;
-        min-height: 30px;
-      }
       .path-separator {
         padding: 0 0.1em;
       }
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Thanks for your review. That makes sense. Maybe a solution would be to impose a max height for desktop or put the styling in a media selector?

jeffcarp:
A media query sounds great!

TimvdLippe:
Sorry for the long delay, I finally had some time again. I limited the sizing to 800px, which `#content` also had.

jeffcarp:
No worries at all. I tried this out locally but for some reason it wasn't working for me. I tried simplifying the query to the following but it still didn't turn the table red when loaded in a <800px width frame. Not sure if this is a web components issue or a media query issue 🤔 

```css
@media (max-width: 800px) {
  table {
    background-color: red;
  }
}
```

TimvdLippe:
@jeffcarp Hm that is really weird. If I add that to the media query, it shows the following:
image
When under 800 pixels wide, the rows are automatically larger with a `::after` in the first `td` of each row. Not sure how to further debug this issue, as it is working for me :thinking: 

Which browser are you using? It is working for me on Chrome and Firefox stable.

jeffcarp:
Apologies again for the delay. Doesn't work for me in Chrome 61.0.3163.100 or 63.0.3230.0 but it does work in FF 55.0.2  `¯\_(ツ)_/¯`

Going to land this. When I deploy if the problem persists I'll look into it and potentially file a Chromium bug.

jeffcarp:
Deployed: https://wpt.fyi/

Weird. It works for me in Chrome when I don't have devtools open. Maybe devtools interferes with the media query?

jeffcarp:
Thanks for your PR and sorry it took so long to land!

TimvdLippe:
@jeffcarp Thanks for merging! It seems to be working just fine on my Chrome 61.0.3163.100 (Official Build) (64-bit) which is the same build as yours.

Issue #107, opened 2017-08-31
Add caveats to about text?


RByers:
Since web devs (and even tech press) may look at the dashboard, we might want to add some caveats about how not to interpret the results.  Eg:
- the test suite has inconsistent quality and coverage - many failures are test bugs
- pass rates are not scores.  A minor issue sometimes causes many test failures, while a major issue may cause only a single failure (or possibly none at all!)
- the suite and dashboard is designed to be useful to browser and standards engineers as part of our shared effort to make the web as a whole higher quality.

Thoughts?

jgraham:
Agreed.

foolip:
That would be great. I wouldn't mind making as prominent as a big red warning on first visit, or however far other would be comfortable going :)

thejohnjansen:
Yeah, given the state of the tests, I'd think a big red warning would help. As we build confidence in the suite and the results posted, then things might change, but (for example) some tests are using non-standard ways of writing to the page (document.body.append()) which may cause tests to fail, even though the browser passes the actual test itself. Until we are able to really get this suite to the point of engineering confidence, I'm wary of giving it too much weight.

jeffcarp:
IMHO I think a big red caveat on all pages is appropriate until items like #98 and what @thejohnjansen mentioned are resolved.

How does this sound as a caveat:

> Data is intended for web platform implementers and is not a reliable view of web platform features. Also note that tested Edge and Safari are not pre-release versions (link to bug).

(please re-mix as you see fit, I'll work on a PR)

jeffcarp:
Uploaded a PR for this: #111 
And deployed the branch to https://caveat-dot-wptdashboard.appspot.com

jeffcarp:
Landed: http://wpt.fyi/

PR's welcome if you think the wording should be different.

RByers:
Looks great, thanks Jeff!

PR #111, opened 2017-08-31
Add caveat on results pages

jeffcarp:
Via discussion in #107.

See it here: https://caveat-dot-wptdashboard.appspot.com

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -93,7 +93,7 @@ limitations under the License.
     </section>
 
     <section class="caveat">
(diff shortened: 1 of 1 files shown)


rwaldron:
👍

jgraham:
I don't know what "reliable view of web-platform features" means. In particular "view" in that sentence is hard to interpret. I assume you mean something like "Data is […] not a useful metric for [evaluating feature completeness|comparing browsers|assessing overall quality]"

jeffcarp:
@jgraham I was hoping "view" would be general enough e.g. to cover the use cases you listed, but I can make the wording more precise.

Since there hasn't been any other feedback on this today I'm going to land it (after tweaking the language as mentioned above).

jgraham:
Thanks, I think that wording is a big improvement.

Issue #90, opened 2017-08-21
webdriver test results wrong


andreastt:
The [WebDriver](https://w3c.github.io/webdriver/webdriver-spec.html) test results shown at http://wpt.fyi/webdriver are obviously wrong. For Chrome it appears to have run, but failed, most of the tests. For Firefox it doesn’t appear to have run any tests at all, but is expected to pass a significant number of them.

bobholt:
Going under the assumption this is an upstream web-platform-tests issue until can research further.

andreastt:
WebDriver tests from WPT run fine in the Firefox CI and upstream using the `wpt` helper.

PR #119, opened 2017-09-20
Better setup documentation and errors

AutomatedTester:
I have added documentation on how to run tests and updated error messages to be more meaningful.

PR #121, opened 2017-09-21
Py2 support

AutomatedTester:
This is based on top of #119.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
         - "node"
 
     - language: python
-      python: 3.4
+      python: 2.7
       install: "pip install -r requirements.txt"
       script:
         - pep8 .
--- a/run/run.py
+++ b/run/run.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 import argparse
-import configparser
+import ConfigParser as configparser
 import gzip
 import json
 import platform as host_platform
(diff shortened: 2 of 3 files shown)


jeffcarp:
Thanks for this. Looks like it needs a rebase but otherwise in general lgtm.

gsnedders:
I'd personally prefer we kept Python 3 support, given it seems not inconceivable we'll move over to Python 3 at some point in the future.

AutomatedTester:
@gsnedders I agree that it would have been nice but ideally we need to make sure that our runners are moved to py3, or at least py2 and py3 support, before doing this.

The current setup for wptdashboard is having both py2 and py3 setup in the same project which seems suboptimal at best.

AutomatedTester:
rebased.

jeffcarp:
Let's take a step back, @automatedtester what is your use case for running run/run.py? The machines it's intended for, the WPTD VMs, have no problem installing both pythons.

jgraham:
The use case is being able to run things locally with the exact same settings as CI in order to debug problems. In this case I think that @AutomatedTester is specifically interested in the tests in `webdriver/`.

AutomatedTester:
@jeffcarp This is because setting up this locally has proven to be near on impossible. Following good practises for python, like using a `virtualenv`, means using 2 version of python unnecessarily hard.

I am trying to fix #90 but I haven't even been able to run the tests yet. I have other setup patches locally to be able to make sure the necessary dependencies are there. Some of the dependencies  only run on Python2.7.

bobholt:
:+1: Running locally should be both possible and easy to do in a virtualenv.

PR #126, opened 2017-09-27
Move to calling wpt helper for running tests. Fixes #122

### LINKED ISSUE #122: wptrunner dependencies for each of the browsers never installed
The dependencies for each of the configurations never get installed for wptrunner. These are the `requirements_<browser>.txt` files. We need them otherwise we get import errors when `run.py` gets to start running them.

I have a patch that I am cleaning up.

AutomatedTester:
Using the `./wpt` command we don't need to worry about dependencies required for running things and it passes things through as if you were calling wptrunner directly.

### DIFF
--- a/run/run.py
+++ b/run/run.py
@@ -132,13 +132,6 @@ def main(platform_id, platform, args, config):
         config['gs_results_bucket'], SUMMARY_PATH
     )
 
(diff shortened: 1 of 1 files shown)


bobholt:
:+1:

PR #133, opened 2017-10-02
Correcting use of write()

AutomatedTester:
This was missed when moving things to python2 only support.


gsnedders:
`json.dumps` seems to return `str` (in both Python 2 and Python 3, because why be consistent), so this LGTM.

PR #134, opened 2017-10-02
Remove the use of Firefox profile in configuration.

AutomatedTester:
In #126 the move to using |wpt run| obsoleted the need for setting the
profile as it wpt makes sure the profile is setup correctly.


foolip:
As I understand it, a profile will be needed for https://github.com/w3c/wptdashboard/issues/125. @JKereliuk, can you confirm?

gsnedders:
@foolip the point is to use the same profile as `./wpt run` does, rather than have another copy here.

foolip:
OK then :)

PR #148, opened 2017-10-10
Fix documentation for installing dependencies

zcorpan:
This is a fixup of 18c9d0016a395324f2eb65f385f5a896c499f2fc where
the file extension was omitted. At least on macOS, it doesn't work.


bobholt:
:+1:

PR #151, opened 2017-10-11
Remove --webdriver-binary

jeffcarp:
Via #testing, we were passing the wrong value to
--webdriver-binary for Geckodriver, causing webdriver tests
in Firefox to fail. Now that `wpt run` supports downloading
a webdriver binary itself, we can remove this option.

cc @AutomatedTester

### DIFF
--- a/run/running.example.ini
+++ b/run/running.example.ini
@@ -1,11 +1,9 @@
 [default]
 build_path = $HOME/wptdbuild
-wpt_path = $HOME/gh/w3c/web-platform-tests
-wptd_path = $HOME/gh/w3c/wptdashboard
+wpt_path = $HOME/web-platform-tests
+wptd_path = $HOME/wptdashboard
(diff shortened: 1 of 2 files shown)


run/running.example.ini
```diff
@@ -1,11 +1,9 @@
 [default]
 build_path = $HOME/wptdbuild
-wpt_path = $HOME/gh/w3c/web-platform-tests
-wptd_path = $HOME/gh/w3c/wptdashboard
+wpt_path = $HOME/web-platform-tests
+wptd_path = $HOME/wptdashboard
```
jeffcarp: This change is unrelated. I always have to change this when setting up a bot. Sneaking in a change to the default here.

AutomatedTester:
We could leave code to handle specific webdriver binaries if people wanted to point to things specifically but that can be done in a follow up.

bobholt:
Agree with @AutomatedTester, but this is good. :+1:

Issue #80, opened 2017-08-15
Use `cp` instead of `rsync` for uploading test result files


jeffcarp:
`gsutil rsync` builds and index before transferring files which is unnecessary. `gsutil cp` is probably a better tool for the job.

PR #153, opened 2017-10-13
Dockerize WPT runs & add Jenkins k8s specs

jeffcarp:
This PR introduces a number of big changes:

- Adds a `Dockerfile` for generating a container capable of running any browser of WPT we currently run
  - It does not bundle a version of FF, it uses `--install-browser`
  - Currently this does not include Chrome
  - The whole running setup is end-to-end tested in Travis CI (FF only for now)
- Adds Kubernetes specs (in `k8s/`) for setting up a Jenkins instance (https://ci.wpt.fyi)

Future work after this lands (I will make bugs for these):

- Get Chrome running in the container (this might involve writing `--install-browser` for Chrome)
- Delete `run/run.py` once Chrome support lands
- Further investigate sharding FF builds

More context around this work is described in this comment: https://github.com/w3c/wptdashboard/issues/164#issuecomment-341172569

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+FROM gcr.io/cloud-solutions-images/jenkins-k8s-slave
+
+USER root
+
+# Expose Sauce Connect port
+EXPOSE 4445
+
+RUN apt-get update
+
+# wpt run dependencies
+RUN apt-get install -y python-pip
+RUN pip install virtualenv
+RUN pip install 'requests==2.13.0'
+
+RUN apt-get install -y git
+
+# Needed for hosts_fixup
+RUN apt-get install -y sudo
+
+# Used for running FF and Chrome
+RUN apt-get install -y xvfb
+
+# Used for running FF
+RUN apt-get install -y \
+    libnss3-tools \
+    libgtk-3-common \
+    libdbus-glib-1-2
+
+RUN mkdir /wptdashboard
+ADD . /wptdashboard
+
+RUN chown -R jenkins:jenkins /wptdashboard
--- a/ci_container.sh
+++ b/ci_container.sh
@@ -1,0 +1,15 @@
+: '
+An end-to-end test of both local configurations (Chrome, FF),
+and remote (Edge, Safari).
+
+This script must pass in order for an image to be pushed
+to production for test runners to pull.
+'
+IMAGE_NAME=wptd-testrun-jenkins
+
+docker build -t "${IMAGE_NAME}" .
+
+docker run \
+  -p 4445:4445 \
+  --entrypoint "/bin/bash" "${IMAGE_NAME}" \
+  /wptdashboard/util/ci_container_inner.sh
--- a/docs/running-ops.md
+++ b/docs/running-ops.md
@@ -1,0 +1,15 @@
+# Running Operations Documentation
+
+## Push a new Docker image to the registry
+
+You must be logged into the `wptdashboard` project in `gcloud`. Be advised
+this is risky as all Jenkins builds pull and use this container.
+
+```sh
+LOCAL_NAME=wptd-testrun-jenkins
+IMAGE_NAME=gcr.io/wptdashboard/$LOCAL_NAME
+
+docker build -t $LOCAL_NAME .
+docker tag $LOCAL_NAME $IMAGE_NAME
+gcloud docker -- push $IMAGE_NAME
+```
--- a/k8s/jenkins.yaml
+++ b/k8s/jenkins.yaml
@@ -1,0 +1,48 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: jenkins
+  namespace: jenkins
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: master
+    spec:
+      containers:
+      - name: master
+        image: jenkins/jenkins:2.67
+        ports:
+        - containerPort: 8080
+        - containerPort: 50000
+        readinessProbe:
+          httpGet:
+            path: /login
+            port: 8080
+          periodSeconds: 10
+          timeoutSeconds: 5
+          successThreshold: 2
+          failureThreshold: 5
+        env:
+        - name: JENKINS_OPTS
+          valueFrom:
+            secretKeyRef:
+              name: jenkins
+              key: options
+        volumeMounts:
+        - mountPath: /var/jenkins_home
+          name: jenkins-home
+        resources:
+          limits:
+            cpu: 500m
+            memory: 1500Mi
+          requests:
+            cpu: 500m
+            memory: 1500Mi
+      volumes:
+      - name: jenkins-home
+        gcePersistentDisk:
+          pdName: jenkins-home
+          fsType: ext4
+          partition: 1
--- a/k8s/lb/ingress.yaml
+++ b/k8s/lb/ingress.yaml
@@ -1,0 +1,13 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: jenkins
+  namespace: jenkins
+  annotations:
+    kubernetes.io/ingress.allow-http: "false"
+spec:
+  tls:
+  - secretName: tls
+  backend:
+    serviceName: jenkins-ui
+    servicePort: 8080
--- a/k8s/options
+++ b/k8s/options
@@ -1,0 +1,1 @@
+--argumentsRealm.passwd.jenkins=CHANGE_ME --argumentsRealm.roles.jenkins=admin
--- a/k8s/service_jenkins.yaml
+++ b/k8s/service_jenkins.yaml
@@ -1,0 +1,27 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: jenkins-ui
+  namespace: jenkins
+spec:
+  type: NodePort
+  selector:
+    app: master
+  ports:
+    - protocol: TCP
+      port: 8080
+      targetPort: 8080
+      name: ui
+kind: Service
+apiVersion: v1
+metadata:
+  name: jenkins-discovery
+  namespace: jenkins
+spec:
+  selector:
+    app: master
+  ports:
+    - protocol: TCP
+      port: 50000
+      targetPort: 50000
+      name: slaves
--- a/run/jenkins.py
+++ b/run/jenkins.py
@@ -116,6 +116,21 @@ def main():
         ]
         if args['run_path']:
             command.insert(3, args['run_path'])
+    elif platform['browser_name'] == "firefox":
+        command = [
+            'xvfb-run', '--auto-servernum',
+            './wpt', 'run',
+            platform['browser_name'],
+            '--install-fonts',
+            '--install-browser',
+            '--yes',
+            '--log-mach=%s' % LOCAL_LOG_FILEPATH,
+            '--log-wptreport=%s' % LOCAL_REPORT_FILEPATH,
+            # temp fix for webRTC
+            '--setpref', 'media.navigator.streams.fake=true',
+        ]
+        if args['run_path']:
+            command.insert(5, args['run_path'])
     else:
         command = [
             'xvfb-run', '--auto-servernum',
--- a/run/jenkins.py
+++ b/run/jenkins.py
@@ -8,20 +8,21 @@ import time
 import re
 
 
+# For making instance-internal network requests for metadata.
+# See: https://cloud.google.com/compute/docs/storing-retrieving-metadata
 METADATA_URL = 'http://metadata.google.internal/computeMetadata/v1'
 
 
 def main():
     # TODO after --install-browser verify browser
-    # version or change platform ID to correct browser version
-    # This can be done by running `wpt install firefox` and then verifying
+    #      version or change platform ID to correct browser version
+    #      This can be done by running `wpt install firefox` and then verifying
     # TODO report OS version when creating TestRun
     # TODO check out correct WPT revision
-
     # TODO convert these back into command line args. They're currently
-    # env vars due to an earlier change. They should be converted back.
-
+    #      env vars due to an earlier change. They should be converted back.
     # TODO break this file up into smaller, unit tested modules.
+    # TODO document how we use metadata
     args = {
         'prod_run': bool(os.environ.get('PROD_RUN', False)),
         'prod_wet_run': bool(os.environ.get('PROD_WET_RUN', False)),
@@ -180,7 +181,7 @@ def main():
     print('==================================================')
     print('Uploading results to gs://%s' % GS_RESULTS_BUCKET)
 
-    # TODO: change this from rsync to cp
+    # TODO(#80): change this from rsync to cp
     command = ['gsutil', '-m', '-h', 'Content-Encoding:gzip',
                'rsync', '-r', args['SHA'], 'gs://wptd/%s' % args['SHA']]
     return_code = subprocess.check_call(command, cwd=args['output_path'])
--- a/run/jenkins.py
+++ b/run/jenkins.py
@@ -1,0 +1,280 @@
+# This replaces run/run.py
+import json
+import gzip
+import os
+import requests
+import subprocess
+import time
+import re
+
+
+METADATA_URL = 'http://metadata.google.internal/computeMetadata/v1'
+
+
+def main():
+    # TODO after --install-browser verify browser
+    # version or change platform ID to correct browser version
+    # This can be done by running `wpt install firefox` and then verifying
+    # TODO report OS version when creating TestRun
+    # TODO check out correct WPT revision
+
+    # TODO convert these back into command line args. They're currently
+    # env vars due to an earlier change. They should be converted back.
+
+    # TODO break this file up into smaller, unit tested modules.
+    args = {
+        'prod_run': bool(os.environ.get('PROD_RUN', False)),
+        'prod_wet_run': bool(os.environ.get('PROD_WET_RUN', False)),
+        'SHA': os.environ.get('WPT_SHA'),
+        'sauce_from_metadata': bool(os.environ.get('SAUCE_FROM_METADATA')),
+        'sauce_key': os.environ.get('SAUCE_KEY', ''),
+        'sauce_user': os.environ.get('SAUCE_USER', ''),
+        'jenkins_job_name': os.environ.get('JOB_NAME'),
+        'wpt_path': os.environ.get('WPT_PATH'),
+        'wptd_path': '/wptdashboard',
+        'output_path': '/wptdout',
+
+        # Passing RUN_PATH will run a specific path in WPT (e.g. html)
+        'run_path': os.environ.get('RUN_PATH', ''),
+    }
+
+    # Arg validation
+    assert args['wpt_path'], '`WPT_PATH` env var required.'
+    assert not args['wpt_path'].endswith('/'), '`WPT_PATH` cannot end with /.'
+    assert len(args['SHA']) == 10, '`WPT_SHA` env var required.'
+
+    if args['prod_run'] or args['prod_wet_run']:
+        print 'Fetching upload_secret from GCE Metadata...'
+
+        url = '%s/project/attributes/upload_secret' % METADATA_URL
+        res = requests.get(url, headers={'Metadata-Flavor': 'Google'})
+        args['upload_secret'] = res.text
+        assert len(args['upload_secret']) == 64, (
+            'Metadata `upload_secret` must exist for prod runs.')
+
+    platform_id, platform = get_and_validate_platform(args['wptd_path'])
+
+    PROD_HOST = 'https://wptdashboard.appspot.com'
+    GS_RESULTS_BUCKET = 'wptd'
+    LOCAL_LOG_FILEPATH = '%s/wptd-testrun.log' % args['output_path']
+    LOCAL_REPORT_FILEPATH = "%s/wptd-%s-%s-report.log" % (
+        args['output_path'], args['SHA'], platform_id
+    )
+    SUMMARY_PATH = '%s/%s-summary.json.gz' % (args['SHA'], platform_id)
+    LOCAL_SUMMARY_GZ_FILEPATH = "%s/%s" % (args['output_path'], SUMMARY_PATH)
+    GS_RESULTS_FILEPATH_BASE = "%s/%s/%s" % (
+        args['output_path'], args['SHA'], platform_id
+    )
+    GS_HTTP_RESULTS_URL = 'https://storage.googleapis.com/%s/%s' % (
+        GS_RESULTS_BUCKET, SUMMARY_PATH
+    )
+
+    SUMMARY_FILENAME = '%s-%s-summary.json.gz' % (args['SHA'], platform_id)
+    SUMMARY_HTTP_URL = 'https://storage.googleapis.com/%s/%s' % (
+        GS_RESULTS_BUCKET, SUMMARY_FILENAME
+    )
+
+    if platform.get('sauce'):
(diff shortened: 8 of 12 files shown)


lukebjerring:
@mdittmer has ramped up on Docker recently - will get him to eyeball this.

ci_container.sh
```diff
@@ -0,0 +1,8 @@
+IMAGE_NAME=wptd-testrun-jenkins
+
+docker build -t $IMAGE_NAME .
```
mdittmer: nit: `"${IMAGE_NAME}"`

ci_container.sh
```diff
@@ -0,0 +1,8 @@
...
+
+docker run \
+  -p 4445:4445 \
+  --entrypoint "/bin/bash" $IMAGE_NAME \
```
mdittmer: ditto

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN apt-get install -y xvfb
+
+# Used for running FF
+RUN apt-get install -y \
+    libnss3-tools \
+    libgtk-3-common \
+    libdbus-glib-1-2
+
+RUN mkdir /wptdashboard
+ADD . /wptdashboard
```
mdittmer: I think our base image should only include dependency installation and possibly some general environment setup (i.e., exclude `ADD` here and `RUN` below). Could we have a `Dockerfile.jenkins` or something for this?

ci_container.sh
mdittmer: May need more flags if we want to build from `Dockerfile.jenkins` or similar.

mdittmer: Can we incorporate a version number here?

docs/running-ops.md
```diff
@@ -0,0 +1,15 @@
+# Running Operations Documentation
+
+## Push a new Docker image to the registry
+
+You must be logged into the `wptdashboard` project in `gcloud`. Be advised
+this is risky as all Jenkins builds pull and use this container.
+
+```sh
+LOCAL_NAME=wptd-testrun-jenkins
```
mdittmer: Version number?

k8s/jenkins.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: extensions/v1beta1
```
mdittmer: Pure ignorance on my part. What does `k8s` stand for?

mdittmer: I'm not qualified to review Jenkins configuration, but could we get some comments in these yaml files to explain what they're for? Running CI for our code? Running WPT? ...?

k8s/lb/ingress.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: extensions/v1beta1
```
mdittmer: Ditto: comments. What's Ingress? (Again, just ignorance on my part, but short explanation and a link or two would help.)

k8s/options
```diff
@@ -0,0 +1 @@
+--argumentsRealm.passwd.jenkins=CHANGE_ME --argumentsRealm.roles.jenkins=admin
```
mdittmer: Is the intention that this gets changed locally? My strategy has generally been to use .md files to document what local configuration needs to be created, and not check it in. If we do want to check it in, maybe checking in `options.example` but actually consuming `options` would be better.

k8s/service_jenkins.yaml
```diff
@@ -0,0 +1,27 @@
+kind: Service
```
mdittmer: Comments plz.

util/ci_container_inner.sh
```diff
@@ -0,0 +1,14 @@
+export WPT_PATH=/web-platform-tests
```
mdittmer: Header comments please :smiley_cat:

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+import subprocess
+import time
+import re
+
+
+METADATA_URL = 'http://metadata.google.internal/computeMetadata/v1'
+
+
+def main():
+    # TODO after --install-browser verify browser
```
mdittmer: With @mattl, @mdittmer, @lukebjerring, @jeffcarp and possibly others now on this project, should we start assigning owners to TODOs?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    # Arg validation
+    assert args['wpt_path'], '`WPT_PATH` env var required.'
+    assert not args['wpt_path'].endswith('/'), '`WPT_PATH` cannot end with /.'
+
+    if args['prod_run'] or args['prod_wet_run']:
+        print 'Fetching upload_secret from GCE Metadata...'
+
+        url = '%s/project/attributes/upload_secret' % METADATA_URL
```
mdittmer: How is this request authenticated?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+        res = requests.get(url, headers={'Metadata-Flavor': 'Google'})
+        args['upload_secret'] = res.text
+        assert len(args['upload_secret']) == 64, (
+            'Metadata `upload_secret` must exist for prod runs.')
```
mdittmer: Is there a common reason this would occur? Is it usually failing to setup something (like app default creds on a gcloud project)? If so, maybe we could have a more informative message.

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+
+    platform_id, platform = get_and_validate_platform(args['wptd_path'])
+
+    PROD_HOST = 'https://wptdashboard.appspot.com'
+    GS_RESULTS_BUCKET = 'wptd'
+    LOCAL_LOG_FILEPATH = '%s/wptd-testrun.log' % args['output_path']
+    LOCAL_REPORT_FILEPATH = "%s/wptd-%s-%s-report.log" % (
+        args['output_path'], args['SHA'], platform_id
```
mdittmer: `args['SHA']` could be `None`, right? Should there be more validation above?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+
+    if platform.get('sauce'):
+        if args['sauce_from_metadata']:
+            print 'Fetching Sauce creds from GCE metadata...'
+
+            for key in ('sauce_user', 'sauce_key'):
+                url = '%s/project/attributes/%s' % (METADATA_URL, key)
+                res = requests.get(url, headers={'Metadata-Flavor': 'Google'})
+                args[key] = res.text
+                assert args[key], 'Metadata key %s is empty.' % key
```
mdittmer: Ditto. Is there an obvious reason this might be the case that we can tell the user?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+
+        assert args['sauce_key'], 'SAUCE_KEY env var required'
+        assert args['sauce_user'], 'SAUCE_USER env var required'
+        SAUCE_TUNNEL_ID = '%s_%s' % (platform_id, int(time.time()))
+
+    assert len(args['SHA']) == 10, 'SHA must a WPT SHA[:10]'
+
+    # Hack because Sauce expects a different name
+    # Maybe just change it in browsers.json?
```
mdittmer: IMHO browsers.json should have a schema that includes an optional `sauce_name`. Code can fallback on `browser_name` when it's not set.

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+            '--sauce-user=%s' % args['sauce_user'],
+            '--sauce-tunnel-id=%s' % SAUCE_TUNNEL_ID,
+            '--no-restart-on-unexpected',
+            '--processes=2',
+            '--run-by-dir=3',
+            '--log-mach=-',
+            '--log-wptreport=%s' % LOCAL_REPORT_FILEPATH,
+            '--install-fonts'
+        ]
+        if args['run_path']:
```
mdittmer: Let's document what this branch does and why/when it's necessary.

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+            'xvfb-run', '--auto-servernum',
+            './wpt', 'run',
+            platform['browser_name'],
+            '--install-fonts',
+            '--install-browser',
+            '--yes',
+            '--log-mach=%s' % LOCAL_LOG_FILEPATH,
+            '--log-wptreport=%s' % LOCAL_REPORT_FILEPATH,
+        ]
+        if args['run_path']:
```
mdittmer: Ditto. What's this do? Why the `3` above and the `5` here?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+            command.insert(5, args['run_path'])
+
+    if args['jenkins_job_name'] and '_of_' in args['jenkins_job_name']:
```
mdittmer: This seems brittle. Maybe a `TODO` to implement a more rigorous interface?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+
+    print('==================================================')
+    print('Finished WPT run')
+    print('Return code from wptrunner: %s' % return_code)
+
+    with open(LOCAL_REPORT_FILEPATH) as f:
+        report = json.load(f)
+
+    assert len(report['results']) > 0, (
+        '0 test results, something went wrong, stopping.')
```
mdittmer: Any hints at where to look for logged errors to diagnose?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+    print('Writing individual result files to local filesystem')
+    for result in report['results']:
+        test_file = result['test']
+        filepath = '%s%s' % (GS_RESULTS_FILEPATH_BASE, test_file)
+        write_gzip_json(filepath, result)
+        print('Wrote file %s' % filepath)
+
+    if not (args['prod_run'] or args['prod_wet_run']):
+        print('==================================================')
+        print('Stopping here (pass PROD_RUN env var to upload results).')
```
mdittmer: nit: `set` rather than `pass`

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+        return
+
+    print('==================================================')
+    print('Uploading results to gs://%s' % GS_RESULTS_BUCKET)
+
+    # TODO: change this from rsync to cp
```
mdittmer: Is there context here? A bug to track? Why are we not already using `cp`?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+    command = ['gsutil', '-m', '-h', 'Content-Encoding:gzip',
+               'rsync', '-r', args['SHA'], 'gs://wptd/%s' % args['SHA']]
+    return_code = subprocess.check_call(command, cwd=args['output_path'])
+    assert return_code == 0
```
mdittmer: No assert message?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+    print('Successfully uploaded!')
+    print('HTTP summary URL: %s' % GS_HTTP_RESULTS_URL)
+
+    print('==================================================')
+    print('Creating new TestRun in the dashboard...')
+    url = '%s/test-runs' % PROD_HOST
```
mdittmer: Is this hitting a data API that runs in AppEngine?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+    )
+    p.communicate(input=patch)
+
+
+def get_and_validate_platform(wptd_path):
+    with open('%s/browsers.json' % wptd_path) as f:
+        browsers = json.load(f)
+
+    platform_id = os.environ['PLATFORM_ID']
+    assert platform_id, 'PLATFORM_ID env var required'
```
mdittmer: Can we list legitimate values here?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+    assert platform_id in browsers, 'PLATFORM_ID not found in browsers.json'
```
mdittmer: ... and here

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+        platform['os_name'], platform['os_version'])
+    )
+
+    p = subprocess.Popen(
+        ['git', 'apply', '-'], cwd=wpt_path, stdin=subprocess.PIPE
+    )
+    p.communicate(input=patch)
+
+
+def get_and_validate_platform(wptd_path):
```
mdittmer: Documentation plz :smile_cat:

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+        browsers = json.load(f)
+
+    platform_id = os.environ['PLATFORM_ID']
+    assert platform_id, 'PLATFORM_ID env var required'
+    assert platform_id in browsers, 'PLATFORM_ID not found in browsers.json'
+    return platform_id, browsers.get(platform_id)
+
+
+def report_to_summary(wpt_report):
+    test_files = {}
```
mdittmer: Ditto: Show me the docs.

mdittmer: Should there be code sharing between this file and `run.py`?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+
+    for result in wpt_report['results']:
+        test_file = result['test']
+        assert test_file not in test_files, (
+            'Assumption that each test_file only shows up once broken!')
```
mdittmer: More context? Which key showed up twice?

run/jenkins.py
```diff
@@ -0,0 +1,282 @@
...
+
+        if result['status'] in ('OK', 'PASS'):
```
mdittmer: Is there a schema somewhere we could lift legitimate values from? (I'm assuming not, but on the off chance there is...)

Dockerfile
jeffcarp: What would the base image be used for other than test runs?

ci_container.sh
jeffcarp: Done. Why is the bare `$IMAGE_NAME` undesirable?

jeffcarp: The image we push for the Jenkins bots to pull needs to have a constant name (otherwise updating the image will require changing the version in Jenkins).

k8s/jenkins.yaml
jeffcarp: Kubernetes

jeffcarp: I'll add a README in this directory. These files are from the [GCP docs on how to set up Jenkins on GKE](https://cloud.google.com/solutions/jenkins-on-container-engine).

k8s/options
jeffcarp: (this file is in the context of the Jenkins GKE guide, see above)

run/jenkins.py
jeffcarp: This is a network-internal request to the GCE metadata store:

> When you make a request to get information from the metadata server, your request and the subsequent metadata response never leaves the physical host running the virtual machine instance.

See https://cloud.google.com/compute/docs/storing-retrieving-metadata#is_metadata_information_secure

run/jenkins.py
jeffcarp: (just some context on this file specifically - `run/jenkins.py` in this PR is a clone of `run/run.py` adapted to run in the container. Once this lands and the container supports all platforms, we can delete `run/run.py`)

jeffcarp: To answer your question: I can make bugs for all of these and we can decide on owners later.

run/jenkins.py
jeffcarp: Added validation, thanks.

run/jenkins.py
jeffcarp: I haven't yet seen this happen. I'll make a note to document how we use metadata somewhere.

run/jenkins.py
jeffcarp: Just to clarify in this case, the user == the maintainer of our Jenkins runs

run/jenkins.py
jeffcarp: That would work. I've been keeping the patch around in wpt.fyi because this problem is an actual bug in wptrunner: https://github.com/w3c/web-platform-tests/issues/6852

I've been meaning to write a fix for that at some point.

run/jenkins.py
jeffcarp: Done

run/jenkins.py
jeffcarp: `wpt run` expects the run path as an ordered argument. It's 3 for Sauce runs because we invoke `['wpt', 'run', $platform-id ... ]` and 5 for Chrome/FF runs because we invoke `['xvfb-run', '--auto-servernum', 'wpt', 'run', $platform-id ... ]`

run/jenkins.py
jeffcarp: This is the beginning of trying to shard runs, but the functionality is not complete, so I'll remove these lines and follow up with a later PR.

run/jenkins.py
jeffcarp: Where the actual logs are? You can find them at URLs like https://ci.wpt.fyi/job/edge-latest-stable/17/console

run/jenkins.py
jeffcarp: Bug: #80 

We are not already using `cp` because I am an idiot

run/jenkins.py
jeffcarp: Done

run/jenkins.py
jeffcarp: This is hitting the actual AE app: `test_runs.go`

run/jenkins.py
jeffcarp: Done

run/jenkins.py
jeffcarp: Done

run/jenkins.py
jeffcarp: Done

jeffcarp: (see above - this file is meant to replace run.py)

run/jenkins.py
jeffcarp: Added a comment.

run/jenkins.py
jeffcarp: A schema for the possible test results? I'm not sure. This line in particular should probably have a discussion bug filed for it.

util/ci_container_inner.sh
jeffcarp: Done

jeffcarp:
Addressed your feedback, thanks for taking a look @mdittmer

Dockerfile
mdittmer: To test our code, to run a dev server for compiling protos, and/or to deploy modules to other environments.

ci_container.sh
mdittmer: If it ever contained spaces or special characters, then the meaning of the command would change.

mdittmer: I think the standard practice for this is to version images, and then use a `:latest` version/tag for use cases such as that. We could also use a version name like `:jenkins` if we want.

k8s/jenkins.yaml
mdittmer: 👍

mdittmer: Excellent. A comment + that link would be useful in the yaml file.

run/jenkins.py
mdittmer: 👍

run/jenkins.py
mdittmer: Right. My thinking is that if someone new to the project is asked to help manage the code or the Jenkins runs, then useful error messages can probably assist ramp-up and save time.

run/jenkins.py
mdittmer: Ah. So we should be able to eliminate the need for this soon. Great. Can you please link to the wpt bug?

run/jenkins.py
mdittmer: SGTM 👍

run/jenkins.py
mdittmer: Perhaps the logs go to the same place as this error message. I assumed that the WPT run occurs in some different context and that its output might not be in the same place this message shows up. Is that false? If so, please disregard.

run/jenkins.py
mdittmer: Glad it's tracked! Thanks.

run/jenkins.py
mdittmer: I see. Is this captured in the design doc somewhere? It felt strange that most of this code "manages its own data" (e.g., pushes JSON straight to GCS), but in one case it delegates to a separate endpoint responsible for managing the data. I was curious about whether there is a technical/design reason for the choice.

run/jenkins.py
mdittmer: Yes, that is what I was asking: does WPT define the data format for its results (including this particular bit of the results).

run/jenkins.py
mdittmer: My apologies; my comment wasn't clear. My intention was that the assertion error message should contain more context (such as the value of `test_file`).

util/ci_container_inner.sh
```diff
@@ -6,6 +6,8 @@ export WPT_PATH=/web-platform-tests
 
 git clone --depth 1 https://github.com/w3c/web-platform-tests $WPT_PATH
 
+sudo chown -R $(whoami):$(whoami) $HOME
```
mdittmer: Prefer `$(id -u $USER):$(id -g $USER)` in case group not named after user.

Dockerfile
jeffcarp: Splitting into multiple Dockerfiles sounds reasonable, but for the sake of simplicity let's start with this one and as the scope is expanded with each PR we can split them up as necessary.

docs/running-ops.md
jeffcarp: I think I addressed this in another comment, but this has to have a specific tag in order to be picked up by Jenkins. As an addition we can push versioned images and then promote them to the prod image name after testing.

run/jenkins.py
jeffcarp: Done

run/jenkins.py
jeffcarp: I'll include this in my documentation of how we use metadata.

run/jenkins.py
jeffcarp: Oops I totally goofed, this is not the hack I was thinking of. Your suggestion is good. I'll make a bug for that.

jeffcarp: #201

run/jenkins.py
jeffcarp: Yep, they go to the same place. We funnel the stdout of the WPT run to the stdout of `run/jenkins.py` by passing `--log-mach=-` to `wpt run`.

run/jenkins.py
jeffcarp: Added the bug number to the TODO

run/jenkins.py
jeffcarp: Yep this is described somewhat in the design doc, https://bit.ly/wptdashboard-design-doc, however the doc is a little out of date. I'll make a bug to add updated running docs to `doc/` in this repo.

util/ci_container_inner.sh
jeffcarp: Done, thanks

jeffcarp:
Ok - I addressed the latest round of feedback. I'm going to go ahead and land this so we have something common to build off of. This is a massive change to our running infra and I think landing this now and following up with incremental improvement PRs is the way to go. :100:

run/jenkins.py
jeffcarp: https://github.com/w3c/wptdashboard/issues/202

run/jenkins.py
jeffcarp: https://github.com/w3c/wptdashboard/issues/203
https://github.com/w3c/wptdashboard/issues/204

PR #156, opened 2017-10-16
Uncomment google-cloud requirement

JKereliuk:
If this isn't done you will get an error if following the current README


gsnedders:
How are we passing tests without this? :(

jeffcarp:
Thanks!

@gsnedders that module is not imported unless you're uploading results in a production run, which isn't functionality that's currently tested

PR #160, opened 2017-10-17
Rename pep8 to pycodestyle

jeffcarp:
It's been renamed: https://github.com/PyCQA/pycodestyle/issues/466


jeffcarp:
Good catch, thanks

PR #165, opened 2017-10-18
Handle the old /static/foo.json results urls

lukebjerring:
Minor tweak which lets the old file-format work a little better in test-file view.


jeffcarp:
Instead of this, could we just update the filenames in /static/? (and at the same time pull more recent results?)

lukebjerring:
Updated the test data in https://github.com/w3c/wptdashboard/pull/167 instead.

PR #167, opened 2017-10-19
Update the populate-dev-data content

lukebjerring:
Uses the SHA[0:10] in the heirarchy too, which is a relied-on hack when creating run urls in the front-end.


jeffcarp:
This is great, thanks!!

Issue #161, opened 2017-10-17
Add endpoint for surfacing 'latest' test run


lukebjerring:
The endpoint could help us determine state-changes for test runs (i.e. detect when a PR is changing the state of a test to be failing when it was passing).
(See https://github.com/w3c/web-platform-tests/issues/7475 for context)

e.g. it would be nice if instead of 
https://storage.googleapis.com/wptd/b12daf6ead/safari-10-macos-10.12-sauce/IndexedDB/abort-in-initial-upgradeneeded.html
we could just call 
https://storage.googleapis.com/wptd/latest/safari-10-macos-10.12-sauce/IndexedDB/abort-in-initial-upgradeneeded.html
where latest works out what to load.

If we can't do so in cloud storage's API, we can add an endpoint somewhere like wpt.fyi/lastest/[test name]

jeffcarp:
Deployed #166, however it looks like we're missing a datastore index:

https://wpt.fyi/latest/chrome-63.0-linux

I haven't double checked but I think adding this to `index.yaml` will resolve the issue:

```
- kind: TestRun
  properties:
  - name: BrowserVersion
  - name: OSName
  - name: CreatedAt
    direction: desc
- kind: TestRun
  properties:
  - name: BrowserVersion
  - name: OSName
  - name: OSVersion
  - name: CreatedAt
    direction: desc
```

lukebjerring:
There's currently a bug with trailing slash handling (fixed in https://github.com/w3c/wptdashboard/pull/168).

However, I'm reconsidering the choice of path for this. I think instead we should be changing the endpoint to be "/json", with ?platform=chrome. /latest is more relevant to the dashboard itself (see 
https://github.com/w3c/wptdashboard/issues/172).

jeffcarp:
I believe this is completed?

PR #169, opened 2017-10-20
Add BrowserVersion, OSName, OSVersion indexes

lukebjerring:
https://github.com/w3c/wptdashboard/issues/161 - Need index for the hyphen-delimited behaviour of that path - worked locally since everything is indexed in the default setup.

e.g. Currently we are getting 500s on the following URL:
https://wpt.fyi/latest/chrome-63.0

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -18,3 +18,24 @@ indexes:
   - name: BrowserName
   - name: CreatedAt
     direction: desc
+- kind: TestRun
+  properties:
+  - name: BrowserName
+  - name: BrowserVersion
+  - name: CreatedAt
+    direction: desc
+- kind: TestRun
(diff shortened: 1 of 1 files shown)


index.yaml
```diff
@@ -16,5 +16,8 @@ indexes:
 - kind: TestRun
   properties:
   - name: BrowserName
+  - name: BrowserVersion
+  - name: OSName
+  - name: OSVersion
```
jeffcarp: Unfortunately I think we need 3 separate indexes now. I'm pretty sure you need an index for each unique set of terms you want to query :cry:

lukebjerring: Done - do we have any concerns for how large these indexes could get?

PR #173, opened 2017-10-24
Rename /latest endpoint to /json

lukebjerring:
Considering /latest might make more sense with respect to https://github.com/w3c/wptdashboard/issues/172, /json is a more appropriate endpoint name for fetching the JSON blob of a given run (for https://github.com/w3c/wptdashboard/issues/161)

/cc @JKereliuk


json_redirect_handler.go
```diff
@@ -29,37 +30,49 @@ import (
...
-// /latest/{browser}[/{test}]
-func latestResultsRedirectHandler(w http.ResponseWriter, r *http.Request) {
-    remainingPath := strings.Replace(r.URL.Path, "/latest/", "", 1)
-    pathPieces := strings.SplitN(remainingPath, "/", 2)
-    if len(pathPieces) > 2 {
+// /json
+//
+// Params:
+//   platform: Browser (and OS) of the run, e.g. "chrome-63.0" or "safari"
+//   (optional) run: SHA[0:10] of the test run
```
jeffcarp: This should mention it defaults to `"latest"`

lukebjerring:
Note that I wanted both commits, unsquashed, because file renames and modifications in the same commit doesn't get handled well in Git.

PR #179, opened 2017-10-24
Firefox webrtc

JKereliuk:
Allows firefox to run some webrtc tests.
For example: https://github.com/w3c/web-platform-tests/blob/master/webrtc/simplecall.html


foolip:
https://wpt.fyi/webrtc is still very red for Firefox, has this taken effect yet?

JKereliuk:
I saw that yesterday as I was heading out :/ going to look at this today

jeffcarp:
Looking at `instance-4`, the revision of wptdashboard it's using is 405943a766c44a5b059de9ab751b716de96865ca from Oct 15. I pulled the code. We can check back tomorrow to see if it solved the problem.

PR #181, opened 2017-10-25
Include SHA in WPT paths

lukebjerring:
Prototype, intended for https://github.com/w3c/wptdashboard/issues/172

### DIFF
--- a/components/test-file-results.html
+++ b/components/test-file-results.html
@@ -81,6 +81,9 @@ limitations under the License.
 
       static get properties () {
         return {
+          sha: {
+            type: String
+          },
           testRuns: {
             type: Array
           },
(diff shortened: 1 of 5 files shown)


jeffcarp:
Thanks for this. Initial thoughts:

- I don't think we should break the current assumption that both paths are equal in wpt.fyi/{`path`} and web-platform-tests/{`path`}
- I think using a URL parameter would be a better choice, e.g. no URL parameter means `latest`, `wpt.fyi/cors?sha=decafbad` means load revision `decafbad`
- Is there a place to access what SHAs are available to load? I guess that's what https://github.com/w3c/wptdashboard/compare/sha-overlay accomplishes

lukebjerring:
Superseded by https://github.com/w3c/wptdashboard/pull/185

PR #183, opened 2017-10-27
Docker + protos

mdittmer:
This PR introduces a `Dockerfile` that loads dependencies for Go, Python, and protobuf support. The intention is that the `sh/` directory will eventually contain several short scripts using Docker as a local build and execution environment. As a "demo" of how this might be used, the PR includes a script for a simple protobuf development server:

```bash
$ /install/docker/on/host/machine
$ cd /path/to/wptdashboard
$ docker build -t wptdashboard:0.1 . # Build Docker image based on ./Dockerfile
$ docker run --rm -it \ # Create and run in interactive mode: instance that will auto-remove when it stops
    -v $(pwd):/wptdashboard \ # Mount (read/write): host path: $(pwd) VM path: /wptdashboard
    --name wptdashboard-instance \ # Name the instance
    wptdashboard:0.1 \ # Use the image just created
    /wptdashboard/sh/watch.sh # Run sh/watch.sh; the development server script
```
If you touch `.proto` files in `protos/`, the development server should log confirmation that it regenerated code from protobuf files.

If useful, we could deploy to test, staging, and production environments using the same Docker image. This would help ensure that issues are easily reproducible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ node_modules
 *.pyc
 run/running.ini
 .DS_Store
+run/protos/*_pb2.py
+bq-schema
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,6 @@ install:
 
 script:
   - sh/ci_script.sh
+
+after_script:
+  - sh/ci_after_script.sh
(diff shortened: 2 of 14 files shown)


mdittmer:
@jeffcarp @gsnedders PTAL

mdittmer:
This code moved to be part of #184

PR #184, opened 2017-10-27
Docker + CI

mdittmer:


sh/ci_after_script.sh
```diff
@@ -0,0 +1,19 @@
...
+set -e
+
+SH_DIR=$(readlink -f $(dirname "$0"))
+source "${SH_DIR}/logging.sh"
+WPTDASHBOARD_DIR=${WPTDASHBOARD_DIR:-$(readlink -f "${SH_DIR}/..")}
+
+cd "${WPTDASHBOARD_DIR}"
+
+info "Stopping and removing docker instance 'wptdashboard-instance'"
+docker stop wptdashboard-instance && docker rm wptdashboard-instance
```
jeffcarp: Is stopping and deleting the instance necessary? The container should presumably stop after the test run, and if it's being run in CI the image management should presumably be taken care of by the runner.

mdittmer: This is mostly for completeness. I don't think I tried leaving this out; I would presume that Travis won't hang on the running image (because the "script" script will run to completion).

mdittmer:
Re: Bazel. I have no problem with using Bazel. I'm not familiar with it, but I can fix that. Are you thinking we would using it for running tests too (as a general automation tool)?

mdittmer:
Obsoleted by #214

PR #185, opened 2017-10-27
Add support for ?run=[SHA]

lukebjerring:
Supercedes https://github.com/w3c/wptdashboard/pull/181 for https://github.com/w3c/wptdashboard/issues/172

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -76,6 +76,16 @@ limitations under the License.
         margin-bottom: 1em;
         border-left: solid 4px #f47477;
       }
+      .info {
+        padding: 0.5em;
+        background-color: #e2e3fd;
+        margin-bottom: 1em;
+        margin-top: 2em;
+        border-left: solid 4px #7477f4;
+      }
+      .info small {
+        float: right;
+      }
 
       @media (max-width: 800px) {
         table tr td:first-child::after {
@@ -97,6 +107,13 @@ limitations under the License.
         <template is="dom-repeat" items="{{ _splitIntoLinkedParts(path) }}" as="part">
           <span class="path-separator">/</span>
           <a href="{{ part.path }}" on-click="_navigate">{{ part.name }}</a>
+        </template>
+
+        <template is="dom-if" if="{{ !isLatest }}">
+          <section class="info">
+            Showing results for run <b>{{ sha }}</b>.
+            <small><a href='?sha=latest'>View latest run</a></small>
+          </section>
         </template>
       </div>
     </section>
@@ -170,6 +187,13 @@ limitations under the License.
             value: '',
             observer: '_queryChanged'
           },
+          sha: {
+            type: String
+          },
+          isLatest: {
+            type: Boolean,
+            computed: '_computeIsLatest(sha)'
+          },
           testRuns: {
             type: Array
           },
@@ -216,6 +240,10 @@ limitations under the License.
         })
 
         this._refreshDisplayedNodes()
+      }
+
+      _computeIsLatest (sha) {
+        return !sha || sha === 'latest'
       }
 
       _computePathIsATestFile (path) {
@@ -323,12 +351,18 @@ limitations under the License.
 
       _navigate (event) {
         event.preventDefault()
-        const path = event.target.pathname.replace(/(.+)(\/)$/, '$1')
+        let path = event.target.pathname.replace(/(.+)(\/)$/, '$1')
         if (path === this.path) {
           return
         }
         this.path = path
         this._refreshDisplayedNodes()
+        if (this.sha && this.sha !== '' && this.sha !== 'latest') {
+          path += path.indexOf('?') < 0
+              ? '?'
+              : '&'
+          path += 'sha=' + this.sha
+        }
         window.history.pushState({}, '', path)
 
         // Send Google Analytics pageview event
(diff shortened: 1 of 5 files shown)


components/wpt-results.html
```diff
@@ -98,6 +104,14 @@
           <span class="path-separator">/</span>
           <a href="{{ part.path }}" on-click="_navigate">{{ part.name }}</a>
         </template>
+
+        <template is="dom-if" if="{{ !isLatest }}">
+          <br /><br />
```
jeffcarp: Can this be replaced with a margin?

components/wpt-results.html
```diff
@@ -98,6 +104,14 @@
...
+          <section class="info">
+            Showing results for run <b>{{ sha }}</b>.
+            <small style='float: right;'><a href='?run=latest'>View latest run</a></small>
```
jeffcarp: Could the inline style be moved up to the stylesheet? It is small and convenient here but I feel like allowing inline styles is a slippery slope.

test_handler.go
```diff
@@ -73,10 +82,32 @@ func testHandler(w http.ResponseWriter, r *http.Request) {
...
         return
     }
 }
+
+func getRunSHA(r *http.Request) (runSHA string, err error) {
+    // Get the SHA for the run being loaded (the first part of the path.)
+    runSHA = "latest"
+    params, err := url.ParseQuery(r.URL.RawQuery)
+    if err != nil {
+        return
```
jeffcarp: This should be returning `nil, err`

test_handler.go
```diff
@@ -73,10 +82,32 @@ func testHandler(w http.ResponseWriter, r *http.Request) {
...
+    }
+
+    runParam := params.Get("run")
```
jeffcarp: If we're calling this `sha` internally what do you think about the URL param also being `sha`?

test_handler.go
```diff
@@ -73,10 +82,32 @@ func testHandler(w http.ResponseWriter, r *http.Request) {
...
+    if runParam != "" && len(runParam) == 10 {
```
jeffcarp: Idea: how about something like

```go
import "regexp"

if !regexp.MatchString("[a-z0-9]{10}", runParam) {
```

(I haven't compiled that though)

components/wpt-results.html
lukebjerring: Done.

components/wpt-results.html
lukebjerring: Done.

test_handler.go
lukebjerring: Can't have nil for a string, and empty/latest are behaviourally equivalent anyway.

test_handler.go
lukebjerring: Done.

test_handler.go
lukebjerring: Done.

test_handler.go
jeffcarp: Ah got it, but if the function will never return an error then maybe its return value shouldn't be `(runSHA string, err error)`? How about just returning `runSHA string` and adding a docstring describing that `""` or `"latest"` means no SHA was found? A unit test would also be great (doesn't need to be in this PR though).

test_handler.go
```diff
@@ -73,10 +83,33 @@ func testHandler(w http.ResponseWriter, r *http.Request) {
...
+    if err != nil {
+        return
+    }
+
+    runParam := params.Get("sha")
+    regex := regexp.MustCompile("[0-f]{10}")
+    if regex.MatchString(runParam) {
+        runSHA = runParam
+    }
+    return
```
jeffcarp: I think this should be `return runSHA`

test_handler.go
```diff
@@ -73,10 +83,33 @@ func testHandler(w http.ResponseWriter, r *http.Request) {
...
+func getRunSHA(r *http.Request) (runSHA string, err error) {
+    // Get the SHA for the run being loaded (the first part of the path.)
+    runSHA = "latest"
+    params, err := url.ParseQuery(r.URL.RawQuery)
+    if err != nil {
+        return
+    }
+
+    runParam := params.Get("sha")
+    regex := regexp.MustCompile("[0-f]{10}")
```
jeffcarp: Cool, I didn't know you could do `0-f`!

lukebjerring: Actually, you can't (includes a wider range of chars). It would be cool if you could, but my bad.

test_handler.go
lukebjerring: that's the same (naked returns in Go... not a fan of them TBH so made it explicit)

test_handler.go
lukebjerring: It does return if it errors (more obvious now with explicit return instead of 'naked return')

lukebjerring: Doc'd and tested.

test_handler_test.go
```diff
@@ -0,0 +1,57 @@
...
+  assert.Nil(t, err)
+  assert.Equal(t, "latest", runSHA)
+}
+
+func TestGetRunSHA_NonSHA_2(t *testing.T) {
+  r := httptest.NewRequest("GET", "http://wpt.fyi/?sha=zapper0123", nil)
+  runSHA, err := GetRunSHA(r)
+  assert.Nil(t, err)
+  assert.Equal(t, "latest", runSHA)
+}
```
jeffcarp: This is great! Thanks!

jeffcarp:
Looks like Travis is having capacity issues, I ran the tests locally and they pass. Will deploy soon. Thanks for this!

PR #186, opened 2017-10-31
Fix #112: update Safari to Safari 11

### LINKED ISSUE #112: Test Safari Technology Preview
I believe the dashboard is testing Safari 10.0. A lot of WPT improvements went into Safari 10.1 and it is very unfortunate those are not reflected on the dashboard.

gsnedders:
Fixes #112.

### DIFF
--- a/browsers.json
+++ b/browsers.json
@@ -56,9 +56,18 @@
         "os_version": "10",
         "sauce": true
     },
-    "safari-10-macos-10.12-sauce": {
+    "safari-11.0-macos-10.12-sauce": {
         "initially_loaded": true,
         "currently_run": true,
+        "browser_name": "safari",
+        "browser_version": "11.0",
+        "os_name": "macos",
+        "os_version": "10.12",
+        "sauce": true
+    },
+    "safari-10-macos-10.12-sauce": {
+        "initially_loaded": false,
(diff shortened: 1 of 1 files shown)


foolip:
Ah, so Sauce has Safari 11 now? @jeffcarp, can you review?

browsers.json
```diff
@@ -56,10 +56,19 @@
         "os_version": "10",
         "sauce": true
     },
-    "safari-10-macos-10.12-sauce": {
+    "safari-11-macos-10.12-sauce": {
```
jeffcarp: I think to futureproof this can we call it `safari-11.0-macos-10.12-sauce`?

browsers.json
```diff
@@ -56,10 +56,19 @@
...
         "initially_loaded": true,
         "currently_run": true,
         "browser_name": "safari",
+        "browser_version": "11",
```
jeffcarp: `11.0` here too

foolip:
@mattl FYI, looks like we'll get Safari 11 with Sauce, but continuing to investigate BrowserStack is still the right path IMHO.

gsnedders:
@jeffcarp done!

jeffcarp:
Thanks! 👍

PR #187, opened 2017-11-01
Add latest prod runs to populate task

lukebjerring:
The SHAs aren't exactly legit, but it does the trick for having the latest runs available and having more than one run exist for each browser.

PR #189, opened 2017-11-02
Revert "Add latest prod runs to populate task"

jeffcarp:
Reverts w3c/wptdashboard#187

cc @lukebjerring reverting this since it doesn't compile:

```
/tmp/tmphxWZ24appengine-go-bin/populate_dev_data_handler.go:75: undefined: http.ErrUseLastResponse
/tmp/tmphxWZ24appengine-go-bin/populate_dev_data_handler.go:86: undefined: http.ErrUseLastResponse
```

### DIFF
--- a/populate_dev_data_handler.go
+++ b/populate_dev_data_handler.go
@@ -15,13 +15,12 @@
 package wptdashboard
 
 import (
-    "fmt"
     "net/http"
     "time"
+    "fmt"
 
     "google.golang.org/appengine"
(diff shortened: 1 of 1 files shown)


lukebjerring:
Interesting. Seems to be an issue with my local version of Go, versus the server's (I actually couldn't get it to compile in local appengine as seen here until I updated ... something.)
The manually-defined error that was there before will do the trick - I'll retry tomorrow.

PR #190, opened 2017-11-02
Add python script for pulling results files

lukebjerring:
Alternative to https://github.com/w3c/wptdashboard/pull/187, this will allow pull from a prod run into a bunch of files used by the repository.


jeffcarp:
Is this meant to replace #192 or live alongside it?

lukebjerring:
Intended to live alongside it - https://github.com/w3c/wptdashboard/pull/192 is for fetching the latest runs with wpt.fyi/json and adding those paths as runs locally (where the data actually lives in prod)

This script is for pulling down a full copy of all the json files for specific run and adding it to the repo, locally (where the data lives locally). Note that this adds somewhere in the order of 100k files to the repo, so I'm no longer convinced it should actually be done. The script is useful for re-using some pieces of code, though.

PR #195, opened 2017-11-03
command line flags and chmod for webrtc

JKereliuk:
Hoping this works on the jenkins build. If not, we can revert :/

cc/ @jeffcarp


jeffcarp:
This is now deployed in our default image (gcr.io/wptdashboard/wptd-testrun-jenkins). Tonight it'll be picked up and hopefully tomorrow we'll see results from it.

PR #198, opened 2017-11-06
Add a tool for diffing runs

lukebjerring:
Simple number diffing for 2 different runs, leveraging the wpt.fyi/json endpoint.

Will be nice to improve this to also have some test path filters, etc.

### DIFF
--- a/util/diff_runs.py
+++ b/util/diff_runs.py
@@ -32,11 +32,11 @@ import math
 import os
 import sys
 import urllib3
-
 from urllib import urlencode
 from typing import List
 
 here = os.path.dirname(__file__)
+
 
 def main():
     args = parse_flags()  # type: argparse.Namespace
--- a/util/diff_runs.py
+++ b/util/diff_runs.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright 2017 Google Inc.
 #
@@ -24,14 +24,16 @@ Example usage:
 """
 
 import argparse
+import certifi
 import inspect
 import json
 import logging
 import math
 import os
 import urllib3
+import sys
 
-from urllib import urlencode
+from urllib.parse import urlencode
 from typing import List
 
 
@@ -44,6 +46,10 @@ def main():
     loggingLevel = getattr(logging, args.log.upper(), None)
     logging.basicConfig(level=loggingLevel)
     logger = logging.getLogger()
+
+    if sys.version_info < (3,):
+        logger.fatal('python3 required')
+        return
 
     differ = RunDiffer(args, logger, Fetcher())
     differ.diff()
@@ -199,7 +205,9 @@ class Fetcher(object):
     '''Placeholder for stubbing 'fetchResults' for the unit tests.'''
 
     def __init__(self):
-        self.pool = urllib3.PoolManager()
+        self.pool = urllib3.PoolManager(
+            cert_reqs='CERT_REQUIRED',
+            ca_certs=certifi.where())
 
     def fetchResults(self, sha, platform):
         '''Fetch a python object representing the test run results JSON for the
@@ -209,7 +217,11 @@ class Fetcher(object):
         encodedArgs = urlencode({'sha': sha, 'platform': platform})
         url = 'https://wpt.fyi/json?' + encodedArgs
 
-        response = self.pool.request('GET', url, redirect=False)
+        try:
+            response = self.pool.request('GET', url, redirect=False)
+        except urllib3.exceptions.SSLError as e:
+            logging.warning('SSL error fetching %s: %s' % (url, e.message))
+            return None
 
         if response.status // 100 != 3:
             logging.warning(
--- a/util/diff_runs.py
+++ b/util/diff_runs.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
 # Copyright 2017 Google Inc.
 #
@@ -30,15 +30,13 @@ import json
 import logging
 import math
 import os
-import urllib3
 import sys
+import urllib3
 
-from urllib.parse import urlencode
+from urllib import urlencode
 from typing import List
 
-
 here = os.path.dirname(__file__)
-
 
 def main():
     args = parse_flags()  # type: argparse.Namespace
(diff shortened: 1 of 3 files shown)


util/diff_runs.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/python3
```
jeffcarp: Sadly we're using python 2 at the moment because WPT infra is all python 2. See https://github.com/w3c/wptdashboard/pull/121

Also: :sob: :sob: :sob: :sob: :sob:

util/diff_runs.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def main():
+    args = parse_flags()  # type: argparse.Namespace
+
+    loggingLevel = getattr(logging, args.log.upper(), None)
+    logging.basicConfig(level=loggingLevel)
+
+    beforeSHA = args.before.sha  # type: str
+    afterSHA = args.after.sha  # type: str
```
jeffcarp: Are these picked up by a tool?

lukebjerring: Yah - it's PEP 484, which is picked up in (some?) python tools + IDEs.

lukebjerring:
Teensy unit tests added.

lukebjerring:
Flicking to @mdittmer since he's comfy with Python and @jeffcarp already has a few Go reviews in transit.

util/requirements.txt
```diff
@@ -0,0 +1,4 @@
+urllib3==1.22
+typing==3.6.2
+protobuf==3.4.0
```
jeffcarp: Is protobuf used in this PR?

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env python3
```
mdittmer: Does this script need Python 3 features?

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
...
+import json
+import logging
+import math
+import os
+import urllib3
+import sys
+
+from urllib.parse import urlencode
+from typing import List
+
```
mdittmer: nit: extra line

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
...
+
+here = os.path.dirname(__file__)
+
```
mdittmer: ditto

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
...
+def main():
+    args = parse_flags()  # type: argparse.Namespace
+
+    loggingLevel = getattr(logging, args.log.upper(), None)
+    logging.basicConfig(level=loggingLevel)
+    logger = logging.getLogger()
+
+    if sys.version_info < (3,):
+        logger.fatal('python3 required')
```
mdittmer: Why?

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
...
+        self.args = args
+        self.logger = logger
+        self.fetcher = fetcher
+
+    def diff(self):
+        beforeSHA = self.args.before.sha  # type: str
+        afterSHA = self.args.after.sha  # type: str
+
+        for i in range(0, len(self.args.after.platforms)):
+            specBefore = '%s@%s' % (self.args.before.platforms[i],
```
mdittmer: We should document somewhere that `--before=foo,bar@SHA --after=baz,quz@SHA` will match `foo:baz` and `bar:quz` for diffing.

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def __init__(self):
+        self.pool = urllib3.PoolManager(
+            cert_reqs='CERT_REQUIRED',
+            ca_certs=certifi.where())
+
+    def fetchResults(self, sha, platform):
+        '''Fetch a python object representing the test run results JSON for the
+        given sha/platform spec. '''
+        # type: (str, str) -> object
```
mdittmer: Can we document the structure of `object`?

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
...
+            if (added > 0):
+                self.logger.info('%d tests ran in %s but not in %s'
+                                 % (added, specAfter, specBefore))
+            if (removed > 0):
+                self.logger.info('%d tests ran in %s but not in %s'
+                                 % (removed, specBefore, specAfter))
+
+
+class Fetcher(object):
+    '''Placeholder for stubbing 'fetchResults' for the unit tests.'''
```
mdittmer: I don't think I understand this comment.

util/diff_runs.py
```diff
@@ -0,0 +1,244 @@
...
+    def fetchResults(self, sha, platform):
+        '''Fetch a python object representing the test run results JSON for the
+        given sha/platform spec. '''
+        # type: (str, str) -> object
+
+        encodedArgs = urlencode({'sha': sha, 'platform': platform})
+        url = 'https://wpt.fyi/json?' + encodedArgs
+
+        try:
+            response = self.pool.request('GET', url, redirect=False)
```
mdittmer: If this is synchronous, what are we getting from the pooling?

util/requirements.txt
```diff
@@ -0,0 +1,4 @@
+certifi==2017.11.5
```
mdittmer: N.B., should be added to `Dockerfile.dev` if #214 lands before this PR.

util/diff_runs.py
lukebjerring: No, it needs python >= 2.7.11. Ubuntu dist is 2.7.6, making it a difficult process to jump through hoops (though, once we have docker running, this becomes redundant)

util/requirements.txt
lukebjerring: As discussed elsewhere, probably best to port Docker to pull from known requirements.txt files for `pip install -r` as it re-uses behaviour that people know/expect.

util/diff_runs.py
lukebjerring: Updated.

util/diff_runs.py
lukebjerring: Done.

util/diff_runs.py
lukebjerring: Done.

util/diff_runs.py
lukebjerring: Done.

util/diff_runs.py
lukebjerring: Done.

util/diff_runs.py
lukebjerring: cert checks.

util/diff_runs.py
lukebjerring: Done.

PR #199, opened 2017-11-06
Serve 'latest' as complete run

lukebjerring:
Resolves https://github.com/w3c/wptdashboard/issues/138

cc @mattl


test_handler.go
```diff
@@ -15,15 +15,16 @@
 package wptdashboard
 
 import (
+    "context"
```
jeffcarp: Is this in the Go standard lib? When I try running this locally it can't find the package `"context"`.

lukebjerring: AFAIK It's https://golang.org/pkg/context/, and it's the context.Context interface that gets returned by appengine.NewContext(request)
https://github.com/w3c/wptdashboard/blob/master/test_handler.go#L43

jeffcarp: Something about both my Go and Python environments is probably majorly screwed up. Just tested this in cloud shell and it works fine.

lukebjerring:
Added the required index - PTAL

test_handler.go
```diff
@@ -110,3 +124,37 @@ func GetBrowserNames() (browserNames []string, err error) {
 	sort.Strings(browserNames)
 	return browserNames, nil
 }
+
+// getLastCompleteRunSHA returns the SHA[0:10] for the most recent run that complete for all of the given browser names.
+func getLastCompleteRunSHA(ctx context.Context, browserNames []string) (sha string, err error) {
+	baseQuery := datastore.
+		NewQuery("TestRun").
+		Order("-CreatedAt").
+		Limit(100).
```
jeffcarp: I believe the results of this query are memcache'd (they would be in Python/ndb, not sure about Go), but we should double check the performance hit of this. I can do a quick check when deploying.

PR #200, opened 2017-11-06
Expose and re-use '/api/run' for a TestRun's JSON

lukebjerring:
We currently dump the runs which are loaded for re-use in the HTML, which isn't re-usable.

This PR adds:
/api/run?browser={platform}&sha={SHA[0:10]}
and
/api/runs?sha={SHA[0:10]}


api_handlers.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2017 Google Inc.
```
mdittmer: I am wholly unqualified to review Go code (at least, right now). @jeffcarp can you plz look at that bit of this PR?

api_handlers.go
```diff
@@ -0,0 +1,138 @@
...
+	"encoding/json"
+	"net/http"
+
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/datastore"
+	"net/url"
+)
+
+// apiTestRunsHandler is responsible for emitting test results JSON for all the runs at the given SHA.
+func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
```
mdittmer: @lukebjerring @jeffcarp are *very* terse variable names a standard idiom in Go?

api_handlers.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		http.Error(w, "Invalid query params", http.StatusBadRequest)
+		return
+	}
+
+	ctx := appengine.NewContext(r)
+	var browserNames []string
+	browserNames, err = GetBrowserNames()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
mdittmer: Here and below. Should we log anything to differentiate between different `500`s?

api_handlers.go
```diff
@@ -0,0 +1,138 @@
...
+	testRunsBytes, err := json.Marshal(testRuns)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Write(testRunsBytes)
+}
+
+// apiTestRunHandler is responsible for emitting the test result JSON for the given browser's run at the given SHA.
```
mdittmer: This comment seems wrong. The data described are in Cloud Storage, but this handler queries Datastore.

api_handlers.go
```diff
@@ -0,0 +1,138 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if len(testRuns) == 0 {
+		http.NotFound(w, r)
+		return
+	}
+
+	testRunsBytes, err := json.Marshal(testRuns[0])
```
mdittmer: Can `len(testRuns) > 1` be `true`? Should we have an assertion?

components/wpt-results.html
```diff
@@ -194,6 +194,11 @@
             type: Boolean,
             computed: '_computeIsLatest(sha)'
           },
+          // ['browser@sha] spec for the runs to load
```
mdittmer: Nit: Missing close `'`?

components/wpt-results.html
```diff
@@ -224,9 +229,22 @@
           this._refreshDisplayedNodes()
         }
 
-        const testFileResults = await Promise.all(this.testRuns.map(testRun => {
-          return this._fetchResults(testRun.results_url)
-        }))
+        const testRuns = await Promise.all(
+          this.testRunSpecs
+            .map(async testRun => {
```
mdittmer: nit: One line `this.testRunSpecs.map(async testRun => {` to save on indents.

components/wpt-results.html
mdittmer: Comment is a little terse. Maybe an example that includes multiple browsers and shas (and a description of what gets loaded).

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "wptdashboard",
   "devDependencies": {
-    "standard": "*",
-    "eslint-plugin-html": "*"
+    "eslint-plugin-html": "*",
+    "standard": "^10.0.3"
```
mdittmer: Why more specific?

api_handlers.go
mdittmer: @jeffcarp do we have a process for Go tests? I don't see any in this PR.

test_handler.go
```diff
@@ -114,4 +86,26 @@ func GetRunSHA(r *http.Request) (runSHA string, err error) {
         runSHA = runParam
     }
     return runSHA, err
+}
+
+// GetBrowserNames loads, parses and returns the set of names of browsers
+// which are to be included (flagged as initially_loaded in the JSON).
+func GetBrowserNames() (browserNames []string, err error) {
```
mdittmer: Should this be copied into memory exactly once? Do we expect the file system of an AppEngine instance to change underneath it without a restart?

api_handlers.go
lukebjerring: Yes, for the unavoibable piping of writer and request. Generally we're more informative with our non-boilerplate naming.

api_handlers.go
lukebjerring: Potentially, though the errors aren't likely

api_handlers.go
lukebjerring: Fixed.

api_handlers.go
lukebjerring: If the data manages to get into a 'duplicated-key' state, we gain little by 500'ing the endpoint (rather than just serving the first one we find).

components/wpt-results.html
lukebjerring: Done.

components/wpt-results.html
lukebjerring: Done.

package.json
lukebjerring: Not sure what auto-updated this. Reverted.

test_handler.go
lukebjerring: Probably not an important optimization (+ cost of development speed locally)

lukebjerring:
+@jeffcarp for the Go stuffs.

api_handlers.go
mdittmer: :+1:

components/wpt-results.html
```diff
@@ -194,11 +194,12 @@
...
             computed: '_computeIsLatest(sha)'
           },
-          // ['browser@sha] spec for the runs to load
+          // ['browser@sha', ...] specs for the runs to load using /api/run
+          // e.g. ['chrome@latest', 'edge@latest', 'firefox@latest', 'safari@latest']
           testRunSpecs: {
             type: Array
           },
-          // JSON blobs for the runs loaded (by their specs)
+          // Fetched + parsed JSON blobs for the runs loaded (by their specs, above)
```
mdittmer: Is the structural type inside the Array small enough to describe it here?

api_handlers.go
mdittmer: I disagree. We should be alerted to the bad data. In particular, if the data have the same `CreatedAt`, then it's feasible for Datastore to deliver us different records each time we hit the endpoint.

lukebjerring: Right, but it's not an HTTP GET request's job to enforce/sanitize the data layer - if we want to detect or avoid those duplications, we should be doing so as part of the schema, or write phase, or both. Besides which, this is sorted by creation date, so we're surfacing the most recent run (which is intended behaviour, if you somehow re-run for a given SHA)

components/wpt-results.html
lukebjerring: Is it standard practice to bloat source-code with comments of the format that the data takes? Shouldn't that (self-)documentation live somewhere else?

mdittmer: Yes, and yes. The trouble is that we don't currently have protos or JSON schema, or `.md` files, or anything for this stuff. If you can use a type name or a link, I'm all for it, but I don't think such types/links exist for these data.

Perhaps it's best to just file (and/or pursue existing?) bugs about structuring the data.

api_handlers.go
jeffcarp: @mdittmer nothing formalized yet. Some basic tests would be great. Here's a good example @lukebjerring landed recently: https://github.com/w3c/wptdashboard/blob/master/json_redirect_handler_test.go

api_handlers.go
```diff
@@ -0,0 +1,146 @@
...
+		return
+	}
+
+	ctx := appengine.NewContext(r)
+
+	query := datastore.
+		NewQuery("TestRun").
+		Order("-CreatedAt").
+		Limit(1).
+		Filter("BrowserName =", browserName)
```
jeffcarp: Do we already have an index for this?

components/wpt-results.html
```diff
@@ -224,9 +230,21 @@
...
+            const url = `/api/run?browser=${browser}&sha=${sha}`
+            const response = await window.fetch(url)
+            return response.json()
+          }))
+
+        const testFileResults = await Promise.all(
+          testRuns
+            .map(async json => {
+              return this._fetchResults(json.results_url)
+            }))
```
jeffcarp: This can be:

```js
const testFileResults = await Promise.all(
  testRuns.map(async json => this._fetchResults(json.results_url))
)
```

test_handler.go
jeffcarp: > Do we expect the file system of an AppEngine instance to change underneath it without a restart?

Looks like AE deploys always restart the VM or create a new one.

> The shutdown process might be triggered by a variety of planned and unplanned events, such as:
> ...
> - You deploy an updated version to the service.

https://cloud.google.com/appengine/docs/standard/go/how-instances-are-managed

In local development too the server process is restarted after a file change.

api_handlers.go
lukebjerring: No, actually (as Revision is added below. Added.

components/wpt-results.html
lukebjerring: Done.

test_handler.go
lukebjerring: Added a TODO for keeping a static copy around.

api_handlers.go
lukebjerring: Will look at re-jigging and adding tests in a follow-up - this PR is already big enough.

jeffcarp:
Deployed this here https://20171109t224019-dot-wptdashboard.appspot.com/

Quickly comparing the two in the profiler, looks like wpt.fyi loads in 3015 ms and that branch loads in 2751 ms. I'm guessing maybe since it only renders once when everything is loaded? wpt.fyi right now feels a bit faster psychologically since it loads the browser table headers first and then the results once they load. For the record I wouldn't be a fan of imposing latency cost for the sake of simplifying code, but I can't argue with the numbers here.

An added benefit of this change is that I believe now you can drop in the `<wpt-results>` element into any website (our GCS bucket has an open CORS policy).

test_handler.go
```diff
@@ -114,4 +86,27 @@ func GetRunSHA(r *http.Request) (runSHA string, err error) {
...
     }
     return runSHA, err
+}
+
+// GetBrowserNames loads, parses and returns the set of names of browsers
+// which are to be included (flagged as initially_loaded in the JSON).
+// TODO(lukebjerring): Persist in memory.
+func GetBrowserNames() (browserNames []string, err error) {
+    var bytes []byte
+    if bytes, err = ioutil.ReadFile("browsers.json"); err != nil {
```
jeffcarp: Not in this PR but we should abstract the loading of `browsers.json` into a utility method since it's done in a few places in our Go code.

api_handlers.go
```diff
@@ -0,0 +1,146 @@
...
+	ctx := appengine.NewContext(r)
+	var browserNames []string
+	browserNames, err = GetBrowserNames()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	var testRuns []TestRun
+	baseQuery := datastore.NewQuery("TestRun").Order("-CreatedAt").Limit(1)
```
jeffcarp: Also here - can we check that we have the necessary indexes?

api_handlers.go
```diff
@@ -0,0 +1,146 @@
...
+	w.Write(testRunsBytes)
+}
+
+// getBrowserParam parses and validates the 'browser' param for the request.
+// It returns "" by default (and in error cases).
+func getBrowserParam(r *http.Request) (browser string, err error) {
+	browser = ""
+	params, err := url.ParseQuery(r.URL.RawQuery)
+	if err != nil {
+		return browser, err
```
jeffcarp: Nit: in some places in this PR the default is returned (naked return), whereas here it's explicit. I don't have a preference other than being consistent.

components/wpt-results.html
```diff
@@ -224,9 +230,19 @@
...
+            const pieces = testRun.split('@')
+            const browser = pieces[0]
+            const sha = pieces.length > 1 ? pieces[1] : 'latest'
+            const url = `/api/run?browser=${browser}&sha=${sha}`
+            const response = await window.fetch(url)
+            return response.json()
+          }))
+
+        const testFileResults = await Promise.all(
+          testRuns.map(async json => { return this._fetchResults(json.results_url) })
```
jeffcarp: The `return` and curly braces are still unnecessary here (unless the linter is making a fuss)

api_handlers.go
jeffcarp: I think the `Limit(1)` will enforce  `testRuns` never being `> 1`, right?

api_handlers.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/datastore"
+	"net/url"
+)
+
+// apiTestRunsHandler is responsible for emitting test-run JSON for all the runs at a given SHA.
+//
+// Params:
+//     sha: SHA[0:10] of the repo when the tests were executed (or 'latest')
```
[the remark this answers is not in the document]
lukebjerring: Done.

test_handler.go
```diff
@@ -114,4 +86,27 @@ func GetRunSHA(r *http.Request) (runSHA string, err error) {
...
+        return nil, err
```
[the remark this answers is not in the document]
lukebjerring: I'll clean up the formatting in a follow-up PR.

test_handler.go
lukebjerring: Will do (think I already do that as part of a PR I'm working on)

components/wpt-results.html
lukebjerring: Done.

api_handlers.go
lukebjerring: We should aim to be explicit - naked is unreadable IMO

lukebjerring: Looks like you're referring to methods that don't have anything to return, though?

api_handlers.go
lukebjerring: Yep - 3rd in the list (https://github.com/w3c/wptdashboard/blob/master/index.yaml#L26)

api_handlers.go
lukebjerring: Correct

api_handlers.go
lukebjerring: Yep (https://github.com/w3c/wptdashboard/blob/master/index.yaml#L16)

api_handlers.go
jeffcarp: Oops, regardless sgtm

Issue #208, opened 2017-11-07
/test-runs is displaying alphabetically, should be chronologically


jeffcarp:
It looks like the runs on https://wpt.fyi/test-runs are alphabetized, they should be in reverse chronological order.

cc @lukebjerring

PR #209, opened 2017-11-07
Order the JSON marshalling of SHAs chronologically

lukebjerring:
Related to https://github.com/w3c/wptdashboard/issues/208, this change makes the output JSON for the test-runs human-readably chronologically ordered (even if subsequent enumeration in JS is not guaranteed to have the same order.)

Note that this may be more important when the JSON blob is exposed through an api endpoint (see https://github.com/w3c/wptdashboard/pull/200)

### DIFF
--- a/test_run_handler.go
+++ b/test_run_handler.go
@@ -15,17 +15,17 @@
 package wptdashboard
 
 import (
+    "bytes"
+    "context"
     "encoding/json"
     "fmt"
     "io/ioutil"
     "net/http"
+    "sort"
     "time"
 
     "google.golang.org/appengine"
     "google.golang.org/appengine/datastore"
-    "sort"
-    "context"
-    "bytes"
 )
 
 func testRunHandler(w http.ResponseWriter, r *http.Request) {
--- a/test_run_handler.go
+++ b/test_run_handler.go
@@ -24,6 +24,8 @@ import (
     "google.golang.org/appengine"
     "google.golang.org/appengine/datastore"
     "sort"
+    "context"
+    "bytes"
 )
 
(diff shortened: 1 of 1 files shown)


test_run_handler.go
```diff
@@ -119,11 +125,21 @@ func handleGet(w http.ResponseWriter, r *http.Request) {
...
+    var orderedSHAs []string
+    orderedSHAs, err = getOrderedSHAs(ctx, createdAtLimit)
+    if err != nil {
+        http.Error(w, err.Error(), http.StatusInternalServerError)
+        return
+    }
+
     // Serialize the data + pipe through the test-runs.html template.
-    testRunsBytes, err := json.Marshal(runs)
+    testRunsBytes, err := marshalInOrder(runs, orderedSHAs)
```
jeffcarp: JavaScript does not guarantee key order for objects - this approach might be easier done in JS.

mdittmer: I think the idea is that a human reader of the API response can see things in the order they expect. @lukebjerring is that right?

lukebjerring: Done in https://github.com/w3c/wptdashboard/pull/211
@mdittmer yes, in theory JSON human-readability makes this change separately valuable even if #211 fixes https://github.com/w3c/wptdashboard/issues/208

jeffcarp:
I don't mean to be a curmudgeon, but if we have a component accepting (and later, an endpoint returning) test runs and the order of those items is important, shouldn't we return a data structure that preserves order, like an array?

lukebjerring:
Yes, we should. Closing this.

PR #211, opened 2017-11-07
Sort runs by created date

lukebjerring:
Resolves https://github.com/w3c/wptdashboard/issues/208

### DIFF
--- a/components/wpt-runs.html
+++ b/components/wpt-runs.html
@@ -125,14 +125,15 @@ limitations under the License.
                 testRuns[results.revision] = {}
               }
               if (!testRuns[results.revision][results.browser_name]) {
-                testRuns[results.revision][results.browser_name] = {
-                  run: results
-                }
+                testRuns[results.revision][results.browser_name] = results
               }
             })
 
         // We flatten into an array of objects so Polymer can deal with them.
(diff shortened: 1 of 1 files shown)


components/wpt-runs.html
```diff
@@ -132,7 +132,9 @@
             })
 
         // We flatten into an array of objects so Polymer can deal with them.
-        const flattened = Object.entries(testRuns).map(entry => ({ sha: entry[0], runs: entry[1] }))
+        const flattened = Object.entries(testRuns)
+            .map(entry => ({ sha: entry[0], runs: entry[1] }))
+            .sort(entry => Object.values(entry.runs)[0].created_date)
```
jeffcarp: Array.prototype.sort takes a function that receives 2 params - item a and item b - and returns `-1`, `0`, or `1`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

lukebjerring:
Fixed - And manually added a range of tests locally to ensure the sorting works correctly.

jeffcarp:
Thanks!

jeffcarp:
Deployed, looks great!

PR #214, opened 2017-11-07
Development inside Docker container

mdittmer:
- `Dockerfile` contains shared base image
- `Dockerfile.jenkins` contains Jenkins-specific setup
- `Dockerfile.dev` contains development server setup
- `util/docker` contains all Docker-related scripts that run in the *host* environment
- `util/docker/inner` contains all Docker-related scripts that run in some *container* environment

New and existing `README.md`s detail environment setup and script usage.


mdittmer:
@lukebjerring FYI. Please give this a try and see whether your Python versioning woes are assuaged.

util/logging.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+GRAY='\033[0;90m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+RED='\033[0;31m'
+NC='\033[0m'
+
+function verbose() {
+  printf "\n${GRAY}[  $(date +'%Y-%m-%d %H:%M:%S')  VERB  ]  $1${NC}\n"
```
lukebjerring: I'd probably drop the trailing newline.

e.g. output is currently

[  2017-11-08 11:18:44  INFO  ]  Creating docker instance for dev server

[  2017-11-08 16:18:44  INFO  ]  SUCCESS: Regen from protos
Setting up watches.  Beware: since -r was given, this may take a while!
Watches established.

mdittmer: Could do. I found that I sometimes didn't get a final newline before my prompt when running scripts, or logging would show up unaligned because it was output on the same line as some dangling output from a sub-process. Both are annoying.

util/docker/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Execute a command in the context of `wptd-dev-instance`.
+
+## su_exec.sh
+
+Execute a command as root in the context of `wptd-dev-instance`.
+
+## jenkins_container.sh
+
+Launch container used by Jenkins CI to perform WPT test runs.
```
jeffcarp: This script is meant to run in Travis CI, not Jenkins CI. It's an e2e test of the test runner so we know that when something lands on master in w3c/wptdashboard, it will (hopefully) work when Jenkins pulls the image and runs it.

util/docker/exec.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# -u $(id -u $USER):$(id -g $USER)  Run as current user and group
+# wptd-dev-instance                 Name of instance (wptdashboard dev server)
+
+EXEC_STR=""
+for ARG in "${@}"; do
+  EXEC_STR+="\"${ARG}\" "
+done
+info "Execute-in-docker:  ${EXEC_STR}"
+docker exec -u $(id -u $USER):$(id -g $USER) wptd-dev-instance "${@}"
```
jeffcarp: Are these exec files necessary? You can open a shell in the containers and debug from there.

util/docker/README.md
mdittmer: Ah. I see.

util/docker/exec.sh
mdittmer: I suppose so. It just makes other scripts that are invoking these scripts more verbose.

util/docker/travis_script.sh
```diff
@@ -0,0 +1,17 @@
...
+"${DOCKER_DIR}/exec.sh" pycodestyle --exclude=*_pb2.py .
+"${DOCKER_DIR}/exec.sh" python -m unittest discover -p '*_test.py'
+"${DOCKER_DIR}/exec.sh" go test -v ./...
+"${DOCKER_DIR}/exec.sh" golint -set_exit_status
+
+# Jenkins infrastructure e2e tests
+docker run \
+  -p 4445:4445 \
+  --entrypoint "/bin/bash" wptd-testrun-jenkins \
+  /wptdashboard/util/docker/inner/travis_jenkins_test.sh
```
jeffcarp: This looks good. It is confusing that we have both Travis and Jenkins for the same project lol

util/docker/travis_script.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -e
+
+DOCKER_DIR=$(dirname "$0")
+
+# WPT Dashboard code tests
+"${DOCKER_DIR}/exec.sh" pycodestyle --exclude=*_pb2.py .
+"${DOCKER_DIR}/exec.sh" python -m unittest discover -p '*_test.py'
+"${DOCKER_DIR}/exec.sh" go test -v ./...
+"${DOCKER_DIR}/exec.sh" golint -set_exit_status
```
jeffcarp: So currently we run the Travis builds in a matrix which allows them to run concurrently. I believe we can preserve this behavior by pre-building the docker container in the Travis run (which is already done in this PR it looks like) and then running each of the above commands as something like:

```
docker run --entrypoint "/bin/bash" wptd-testrun-jenkins pycodestyle --exclude=*_pb2.py
# or
docker run --entrypoint "/bin/bash" wptd-testrun-jenkins python -m unittest discover -p '*_test.py'
```

What do you think?

mdittmer: That sounds good, but I'm not sure I know how to configure using a matrix. My understanding is that `docker run` is a blocking call, so simply doing what you describe inside `travis_script.sh` will not speed up test runs. Is that right?

Dockerfile.dev
```diff
@@ -0,0 +1,85 @@
...
+    core \
+    gsutil \
+    app-engine-python \
+    kubectl
+
+# Put wptdashboard code in GOPATH
+RUN ln -s "/wptdashboard" "/go/src/wptdashboard"
+
+# Drop dev environment into source path
+WORKDIR /wptdashboard
```
jeffcarp: I'm kind of confused about how these files are split up. If the Jenkins runner instances need to send results to BQ using protos, don't they also need the above proto-related things and gcloud installed?

util/docker/inner/travis_jenkins_test.sh
```diff
@@ -1,6 +1,6 @@
 : '
-This file runs inside the testrun container
-in the context of CI.
+This file runs inside a WPT testrun container
+in the context of Travis CI.
```
jeffcarp: 👍

util/docker/inner/watch.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
jeffcarp: I'm a little confused about the directory structure - `util/docker/inner` makes me think that a container is running inside another container. What about:

- `util/docker-dev/` for the server
- `util/docker-jenkins/` for the test runner

Dockerfile.dev
mdittmer: Aha! Yes, they will, eventually. Do you want me to upstream more of the installation in to `.base` now or later?

util/docker-jenkins/inner/run.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# This file runs inside a WPT testrun container
+# in the context of Travis CI.
```
jeffcarp: Since this file is for Travis CI-only it shouldn't be named something as general as `run.sh` - how about `travis_ci_run.sh`

util/docker-jenkins/run.sh
```diff
@@ -0,0 +1,20 @@
+: '
```
jeffcarp: Same for this file

README.md
```diff
@@ -8,38 +8,46 @@ It consists of 3 parts:
...
 
-## Running the tests
+## Running locally
 
-We run the tests with a Python script [`run/run.py`](run/run.py) which is a thin wrapper around WPT's [`wpt run`](https://github.com/w3c/web-platform-tests/#running-tests-automatically). If you're triaging test failures, use `wpt run`.
+In one terminal, start the web server:
 
-### Setup
+```sh
+./util/docker-dev/web_server.sh
```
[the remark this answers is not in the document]
mdittmer: I'd rather not. All the other make targets assume they are running *inside* the development environment (though the paths can be configured to work in a host environment too).

README.md
```diff
@@ -8,38 +8,46 @@ It consists of 3 parts:
...
+```
 
-You'll need to make sure that you have Python 2.7 installed. It is recommended that you setup a [virtualenv](https://virtualenv.pypa.io/en/stable/). When you have activated your `virtualenv`, install the dependencies:
+This will build dependencies and start the Google Cloud development server
```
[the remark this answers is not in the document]
mdittmer: Done.

util/docker-jenkins/inner/run.sh
mdittmer: Done.

util/docker-jenkins/run.sh
mdittmer: Done.

PR #221, opened 2017-11-08
Add test paths filtering to diff tool

lukebjerring:
Allows specific path prefixes, which is useful for comparing results of 2 runs for a given folder.
e.g. `util/diff_runs.py --before=chrome@latest --after=safari@latest /css/css-images/`

### DIFF
--- a/util/diff_runs_test.py
+++ b/util/diff_runs_test.py
@@ -26,6 +26,8 @@ class DiffRunTestCase(unittest.TestCase):
 
     def setUp(self):
         self.mock_args = mock.Mock(spec=argparse.Namespace)
+        self.mock_args.tests = []
+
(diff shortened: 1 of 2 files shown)


util/diff_runs_test.py
```diff
@@ -78,6 +80,28 @@ def results(sha, platform):
         self.assertIn('1 differences', logged)
         self.assertIn('1 tests', logged)
 
+    def test_cull_ignored_tests_dir(self):
```
mdittmer: Can we add tests for removes-all and removes-none?

lukebjerring: Removes none is done as 'test_no_difference', so just test_removes_all added.

PR #222, opened 2017-11-09
Produce a (sorted) TestRun array instead of a map

lukebjerring:


test_run_handler.go
```diff
@@ -138,4 +134,11 @@ func handleGet(w http.ResponseWriter, r *http.Request) {
...
     }
-}
\ No newline at end of file
+}
+
+// ByCreatedAtDesc implements sort.Interface for []TestRun based on the CreatedAt field.
+type ByCreatedAtDesc []TestRun
+
+func (a ByCreatedAtDesc) Len() int           { return len(a) }
+func (a ByCreatedAtDesc) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
```
mdittmer: This feels wrong; why the tautology in the middle?

lukebjerring: it's actually "a[i] = a[j] and a[j] = a[i]", but at the same time. Saves the normal placeholder:

    temp := a[i]
    a[i] = a[j]
    a[j] = temp

Note that this is copy-pasted from the canonical Go example for sort: https://golang.org/pkg/sort/

PR #224, opened 2017-11-10
gofmt all files

lukebjerring:
Done with the command:

    find . -name \*.go -print | xargs gofmt -w


lukebjerring:
@jeffcarp FYI go recommends use of _tabs_, not spaces. 

"I do not approve this message" - lukebjerring

PR #226, opened 2017-11-10
Filter non-200 fetches

lukebjerring:
The test_handler.go produces a spec list for a given sha, and doesn't actually check whether each browser exists for the given sha (which is determined by the /api/run returning a 404).

Fixes https://github.com/w3c/wptdashboard/issues/232

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -237,7 +237,7 @@ limitations under the License.
             const sha = pieces.length > 1 ? pieces[1] : 'latest'
             const url = `/api/run?browser=${browser}&sha=${sha}`
             const response = await window.fetch(url)
-            if (response.status != 200) {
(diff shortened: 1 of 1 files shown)


components/wpt-results.html
```diff
@@ -230,15 +230,19 @@
...
             const browser = pieces[0]
             const sha = pieces.length > 1 ? pieces[1] : 'latest'
             const url = `/api/run?browser=${browser}&sha=${sha}`
             const response = await window.fetch(url)
+            if (response.status !== 200) {
+              return Promise.resolve()
+            }
             return response.json()
           }))
+        const testRuns = fetches.filter(e => e)  // Filter unresolved fetches.
```
jeffcarp: This can also be `fetches.filter(Boolean)`

PR #228, opened 2017-11-10
Run diff prototype

lukebjerring:
Running at https://20171122t141813-dot-wptdashboard.appspot.com/

Example: Chrome 7a0cf8ade7 vs 6ed1ba18d0 (Nov 2nd vs Nov 4th)
https://20171122t141813-dot-wptdashboard.appspot.com/api/diff?before=chrome@7a0cf8ade7&after=chrome@6ed1ba18d0

--- 

This PR adds `/api/diff?before=platform@sha&after=platform@sha` endpoint, for (dangerously simplified) summaries of the diff counts between 2 runs.

Produces the same JSON format as `/results?browser={platform}&sha={sha}`, only instead of

    { "foo": [pass, total], "bar": ... }

it produces

    { "foo": [pass_delta, max(total_tests)], ... }

And omits anything with the same counts. This format can* then be piped into `<wpt-results>`

\* Not yet prototyped

/cc @jeffcarp


foolip:
I can't see "summary" in the diff, but this is based on the summary pass/fail numbers alone, right? Would anything much need to change in order to use the full pass/fail data once that's available in a more convenient format than tens of thousands of individual files?

lukebjerring:
No, should be straightforward to diff all the data once it's unsharded.
This current state does not include subtests.

api_handlers.go
```diff
@@ -257,3 +257,52 @@ func getLastCompleteRunSHA(ctx context.Context, browserNames []string) (sha stri
...
+	var afterJSON map[string][]int
+	if afterJSON, err = fetchRunResultsJSONForParam(ctx, r, specAfter); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	} else if afterJSON == nil {
+		http.Error(w, specAfter+" not found", http.StatusNotFound)
+		return
+	}
+
+	diffJSON := diffResults(beforeJSON, afterJSON)
```
foolip: My first time reading go code. If I'm reading this right, almost everything up to this point is error handling :)

run_diff.go
```diff
@@ -0,0 +1,142 @@
...
+		return platformAtRevision, errors.New("invalid platform@revision spec: " + spec)
+	}
+	platformAtRevision.Platform = pieces[0]
+	if len(pieces) < 2 {
+		// No @ is assumed to be the platform only.
+		platformAtRevision.Revision = "latest"
+	} else {
+		platformAtRevision.Revision = pieces[1]
+	}
+	if IsBrowserName(platformAtRevision.Platform) {
```
foolip: Can you add a comment about this? Seems curious to check if the platform is a browser.

run_diff.go
```diff
@@ -0,0 +1,142 @@
...
+
+func fetchRunForSpec(ctx context.Context, revision platformAtRevision) (TestRun, error) {
+	baseQuery := datastore.
+		NewQuery("TestRun").
+		Order("-CreatedAt").
+		Limit(1)
+
+	var results []TestRun
+	query := baseQuery.
+		Filter("BrowserName =", revision.Platform)
```
foolip: This looks funny in the same way, so now I'm thinking we're just using browser and platform as synonyms in these parts. Is that the case?

run_diff.go
```diff
@@ -0,0 +1,142 @@
...
+	if _, err := query.GetAll(ctx, &results); err != nil {
+		return TestRun{}, err
+	}
+	if len(results) < 1 {
+		return TestRun{}, nil
+	}
+	return results[0], nil
+}
+
+func fetchRunResultsJSON(ctx context.Context, r *http.Request, run TestRun) (results map[string][]int, err error) {
```
foolip: Can you sprinkle some TODOs around here about not fetching lots of little files, and linking to the issue for that?

run_diff.go
```diff
@@ -0,0 +1,142 @@
...
+	client := urlfetch.Client(ctx)
+	url := run.ResultsURL
+	if strings.Index(url, "/") == 0 {
+		reqURL := *r.URL
+		reqURL.Path = url
+	}
+	var resp *http.Response
+	if resp, err = client.Get(url); err != nil {
```
foolip: If this is where network errors end up, it seems somewhat likely that we'll see some. Will the client see a HTTP 500 then? If it's not very unlikely to happen, seems like some retry logic will be needed in the Travis integration?

run_diff.go
```diff
@@ -0,0 +1,142 @@
...
+	if body, err = ioutil.ReadAll(resp.Body); err != nil {
+		return nil, err
+	}
+	if err = json.Unmarshal(body, &results); err != nil {
+		return nil, err
+	}
+	return results, nil
+}
+
+func diffResults(before map[string][]int, after map[string][]int) map[string][]int {
```
foolip: @mdittmer, here are a bunch of loops over test results, this is the shape of things I was imagining would be involved in computing metrics. Just FYI, let's keep going in the design doc.

run_diff.go
```diff
@@ -0,0 +1,142 @@
...
+	diff := make(map[string][]int)
+	for test, resultsBefore := range before {
+		if resultsAfter, ok := after[test]; !ok {
+			// Missing? Then N / N tests are 'different'
```
foolip: Do we have any tests in the code base yet? This is a bit of code that would easier to review the tests for than convince oneself about correctness in edge cases by executing in head.

run_diff.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, err
+	}
+	if err = json.Unmarshal(body, &results); err != nil {
+		return nil, err
+	}
+	return results, nil
+}
+
+func diffResults(before map[string][]int, after map[string][]int) map[string][]int {
+	diff := make(map[string][]int)
```
foolip: Can you add a comment about what the 2-tuple int represents here? I think "number of tests different" and "total number of tests known"?

util.go
```diff
@@ -61,3 +61,35 @@ func GetBrowserNames() ([]string, error) {
...
+		return false
+	}
+	var browsers []string
+	var err error
+	if browsers, err = GetBrowserNames(); err != nil {
+		return false
+	}
+	for _, browser := range browsers {
+		if browser == name {
+			return true
```
foolip: Is there no array.contains(thing) in Go?

util.go
```diff
@@ -61,3 +61,35 @@ func GetBrowserNames() ([]string, error) {
...
+		}
+	}
+	return false
+}
+
+func abs(x int) int {
```
foolip: Don't know go, but if https://golang.org/pkg/math/ is part of the standard library, is it most idiomatic to use that or roll your own?

api_handlers.go
lukebjerring: Yeah, go's error handling is .. verbose.

run_diff.go
lukebjerring: Platform is actually browser[-version[-os[-version]]], in this case enforced as only the browser.

run_diff.go
lukebjerring: TODO's added.

run_diff.go
lukebjerring: I don't think we should open the issue until this code lands (issues against branches feels dirty).

run_diff.go
lukebjerring: Yes, 500, idk about the likelyhood, not going to overengineer until we see an actual problem (travis makes a HUGE number of network fetches for pip etc anyway).

run_diff.go
lukebjerring: Done.

run_diff.go
lukebjerring: Will add some tests shortly and push them into this PR.

util.go
lukebjerring: no.

util.go
lukebjerring: math lib works with floats, not ints, arguing that ints are 'trivial to roll your own'.

(I don't like Go in that regard.)

util.go
foolip: Weird :)

run_diff.go
foolip: Sure, I suppose that this will only make a handful of requests, I was thinking about what @mdittmer told me about in code where he fetched 10k files or so. If we expected that kind of load some precautions might be in order.

run_diff_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func TestDiffResults_Added(t *testing.T) {
+	// One new test, all passing
+	assertDelta(t, []int {1, 1}, []int {2, 2}, []int {1, 2})
+
+	// One new test, all failing
+	assertDelta(t, []int {0, 1}, []int {0, 2}, []int {1, 2})
+
+	// One new test, new test passing
+	assertDelta(t, []int {0, 1}, []int {1, 2}, []int {1, 2})
```
foolip: I I'm reading this right, the numbers returned as the same when a new test is pass and when it's failing, is that right? That would make it impossible to notify about new failing tests but not new passing tests, which I think we should.

lukebjerring: Yeah, currently the case. Removing the abs call would let you see negative values; however, this API will be extended to take a filters param, and you'd filter to only newly-failing in the fetch which is used for notifying.

foolip: Sounds like a plan!

PR #236, opened 2017-11-13
Revert wpt-results to supporting test-runs being inlined.

lukebjerring:
Continues to use the resources from /api/run, but allows the component to have the runs specified as inline JSON (as was the previous impl), which will be needed for re-using the component.

### DIFF
--- a/components/wpt-results.html
+++ b/components/wpt-results.html
@@ -194,12 +194,12 @@ limitations under the License.
             type: Boolean,
             computed: '_computeIsLatest(sha)'
           },
-          // ['browser@sha', ...] specs for the runs to load using /api/run
-          // e.g. ['chrome@latest', 'edge@latest', 'firefox@latest', 'safari@latest']
-          testRunSpecs: {
+          // URLs to load to get the TestRun JSON
+          testRunResources: {
             type: Array
           },
-          // Fetched + parsed JSON blobs for the runs loaded (by their specs, above)
+          // TestRuns JSON objects, either fetched + parsed (from their source URLs, above),
+          // or initially dumped inline in the test-runs attribute.
           testRuns: {
             type: Array
           },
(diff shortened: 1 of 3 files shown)


components/wpt-results.html
```diff
@@ -194,12 +194,12 @@
             type: Boolean,
             computed: '_computeIsLatest(sha)'
           },
-          // ['browser@sha', ...] specs for the runs to load using /api/run
-          // e.g. ['chrome@latest', 'edge@latest', 'firefox@latest', 'safari@latest']
-          testRunSpecs: {
+          // URLs to load to get the TestRun JSON
+          testRunResources: {
```
lukebjerring: Fun fact - if this is called testRunResourceUrls instead, polymer auto-magically escapes the already-escaped string as a URL, which breaks the unmarshalling...

mdittmer: Thanks for "helping" Polymer.

components/wpt-results.html
```diff
@@ -230,18 +230,19 @@
...
-        const testRuns = await Promise.all(
-          this.testRunSpecs.map(async testRun => {
-            const pieces = testRun.split('@')
-            const browser = pieces[0]
-            const sha = pieces.length > 1 ? pieces[1] : 'latest'
-            const url = `/api/run?browser=${browser}&sha=${sha}`
-            const response = await window.fetch(url)
-            return response.json()
-          }))
+        // Fetch the TestRun JSON if it's not dumped inline.
```
mdittmer: This should go through quickly right now because the site is broken. However, is there a plan to avoid the "use this strategy, unless it doesn't work, then use that strategy" here?

lukebjerring: It's not 'if it doesn't work' - it's essentially an alternative ctor for the Polymer element.

lukebjerring:
Build 'failure' is actually an eslint error/issue - merging anyway.

Issue #235, opened 2017-11-13
Test result pages break after working the first time


mdittmer:
Example:

I followed links to get to https://wpt.fyi/2dcontext/fill-and-stroke-styles/2d.fillStyle.parse.css-color-4-hsl-2.html

I saw the results the first time I loaded the page. When I refresh the page I see no results.

@lukebjerring could this be related to the new endpoints you've been integrating?

foolip:
I saw this too when loading https://wpt.fyi/subresource-integrity/subresource-integrity.sub.html and then reloading.

Assigning to @mattl to investigate what's going on.

foolip:
Actually, in a 1:1 with @mdittmer right now and he thinks it's likely that this is related to @lukebjerring's changes, so assigning to Luke first.

lukebjerring:
Caused by the intermediate TestRun JSON fetch not being triggered when loading an individual result directly.

lukebjerring:
Fixed via https://github.com/w3c/wptdashboard/pull/237

foolip:
Confirmed to be working! @lukebjerring, I guess this means that we're now auto-deploying changes? (I expected it to not work until @jeffcarp got to work.)

lukebjerring:
We're working on documenting and sharing the deployment process, as don't
want to overload @jeffcarp (nor bottleneck).

I did this deployment in particular.

PR #237, opened 2017-11-13
Add shared results polymer base class

lukebjerring:
Fix for https://github.com/w3c/wptdashboard/issues/235, which is caused when the reload of an individual-results page does not trigger the (newly added) api fetch for the TestRun JSON.

### DIFF
--- a/components/results.html
+++ b/components/results.html
@@ -1,0 +1,56 @@
+<!--
+Copyright 2017 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


components/results.html
```diff
@@ -0,0 +1,57 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<link rel="import" href="../bower_components/polymer/polymer-element.html">
+
+<dom-module id="wpt-results-base">
+    <script>
+      class ResultsBase extends window.Polymer.Element {
```
jeffcarp: Can you add a docstring about what the base class is responsible for?

lukebjerring: Done.

lukebjerring:
Merging in spite of "broken build" which is actually an eslint bug/issue

PR #240, opened 2017-11-14
Revert https://github.com/w3c/wptdashboard/pull/199

lukebjerring:
SHA order seems incorrect and results are showing from Sept.


mattl:
Reverted!

Issue #241, opened 2017-11-14
/tasks/populate-dev-data shouldn't be accessible in prod


lukebjerring:
Currently that endpoint, which is a write-on-GET, is freely accessible. It should be ported to a local python script and removed.

/cc @mdittmer @jeffcarp

PR #242, opened 2017-11-14
Add script for copying prod run

lukebjerring:
Note that this should not be merged until https://github.com/w3c/wptdashboard/issues/241 is resolved (as it includes a local upload-token setup hack to get this to work, for completeness).


test_run_handler.go
```diff
@@ -37,19 +38,25 @@ func testRunHandler(w http.ResponseWriter, r *http.Request) {
 
 func handlePost(w http.ResponseWriter, r *http.Request) {
 	ctx := appengine.NewContext(r)
-	var err error
 
 	// Fetch pre-uploaded Token entity.
-	suppliedSecret := r.URL.Query().Get("secret")
+  var queryParams url.Values
```
rwaldron: Nit: Strange indentation here?

lukebjerring: Fixed - I'm not sure which editor is doing that, I bounce between pycharm/gogland/sublime/atom

lukebjerring:
Reverted the secret-adding part, as this is now covered by a python script in https://github.com/w3c/wptdashboard/pull/248 which is safe.

test_run_handler.go
```diff
@@ -67,16 +74,19 @@ func handlePost(w http.ResponseWriter, r *http.Request) {
 	}
 	testRun.CreatedAt = time.Now()
 
-	fmt.Fprintf(w, "got... %v", testRun)
```
lukebjerring: Note that this output is (was) a bit _debuggy_. The output is now just the JSON blob as stored in the DB, which is a pretty standard behaviour in a REST API.

Issue #118, opened 2017-09-19
CLA no longer applicable?


jgraham:
I think since the project moved under w3c the requirement to sign the Google CLA is perhaps no longer applicable (and if that isn't the case I would like to figure out how we can make it so). Given that references to the CLA in the documentation should be removed.

foolip:
I'll have to ask for advice and report back.

foolip:
Answer is "signing the Google CLA is no longer applicable, and can be removed. This should probably be updated to point to W3C's CLA, but this is really up to the w3c".

@plehegar @sideshowbarker, who should we talk to about this?

RByers:
Perhaps we should update the LICENSE and any copyright notices to officially transfer the copyright to the W3C?  I assume W3C folks will want to use the 3-clause BSD license?  That should be fine from Google's perspective.

sideshowbarker:
From the W3C side, we should get @wseltzer to look and consider what if anything needs to be done here

wseltzer:
Thanks!  W3C usually uses the [W3C Software and Document License](https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document) , a version of the MIT license. Would the original authors be willing to re-license? We'd then include the relevant texts from https://github.com/w3c/licenses

RByers:
Other WPT tools (eg. [wptrunner](https://github.com/w3c/wptrunner/blob/master/LICENSE)) are using the 3-Clause BSD license.  Google is happy to relicense under that license (since it's nearly identical to our own Chromium BSD license).  If there's some reason to prefer the W3C SDL then I believe we'd need to check to see if there are any concerns from our legal team.

plehegar:
Note that it would make sense to use the 3-Clause BSD license. WPT is using it in https://github.com/w3c/web-platform-tests/blob/master/LICENSE.md (which probably explains why the other WPT tools are using it).

foolip:
I'll take that to mean that https://github.com/w3c/wptrunner/blob/master/LICENSE will be OK then.

What about the CLA itself? Do we just delete it, or is something needed that makes it such that contributions are also licensed appropriate?

PR #243, opened 2017-11-14
Remove the CLA section of CONTRIBUTING.md

foolip:
Per https://github.com/w3c/wptdashboard/issues/118#issuecomment-330689900

The guidance around copyright headers is still in need of updating.


mattl:
This looks good to me. I personally don't think we should worry too much about 3-clause BSD vs Expat license but @wseltzer knows best :)

PR #244, opened 2017-11-14
Update the LICENSE to the W3C 3-clause BSD License

foolip:
Per https://github.com/w3c/wptdashboard/issues/118#issuecomment-332518089

The LICENSE file is copied verbatim from the wptrunner repo, and like in
the web-platform-tests repo, each individual file doesn't get a license
header.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,6 @@
-# Copyright 2017 Google Inc.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
+# Copyright 2017 The WPT Dashboard Project. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
 
 sudo: required
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,17 +28,7 @@ npm test
 All source files (including `.js`, `.go`, `.html`, `.css`) must begin with a comment of the below header:
 
 ```go
-// Copyright {YEAR} Google Inc.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
+// Copyright {YEAR} The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
 ```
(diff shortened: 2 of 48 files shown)


mattl:
We should keep a license header on the files, even with the new license.

foolip:
What should it look like? I'm not a fan of a sometimes-growing list of names and dates, but am a fan of the Chromium style:
```
# Copyright 2014 The Chromium Authors. All rights reserved.
# Use of this source code is governed by a BSD-style license that can be
# found in the LICENSE file.
```

If that's fine, then we'd need to create an AUTHORS file, and also pick something instead of "Chromium", like "web-platform-tests dashboard"?

mattl:
Generally for BSD licensed files I've seen the full license included, especially as there are some confusing variations of the license. So: 

 # Copyright (c) 2017 Web Platform Tests Project <http://wpt.fyi>
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
 #
 # 1. Redistributions of source code must retain the above copyright notice,
 #    this list of conditions and the following disclaimer.
 # 2. Redistributions in binary form must reproduce the above copyright
 #    notice, this list of conditions and the following disclaimer in the
 #    documentation and/or other materials provided with the distribution.
 # 3. Neither the name of mosquitto nor the names of its
 #    contributors may be used to endorse or promote products derived from
 #    this software without specific prior written permission.
 #
 # THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 # AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 # IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
 # ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
 # LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
 # CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
 # SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
 # INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
 # CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.

foolip:
I'd really like to not have such a big block in all files. No per-file license at all or the minimal 3-liner would be my preference. @wseltzer, can you advise on what the minimum thing we can do is?

mattl:
I’d like to see something on each file, so given the choice of nothing or a short three line notice, we should go with that if we can @wseltzer

rwaldron:
2 ¢: We are trying to get rid of "every file has a license header" in Test262, because it's really awful to maintain in the long term.

mattl:
The reason for a header (even if its a short one) is that files often get separated from their original project, and down the line will make things difficult to adopt if they have unknown license terms.

I think the 3 line header is a good compromise between a 9000 line header and no head.

foolip:
The thing to decide here is what the first line would be. Does anyone mind "Copyright 2014 web-platform-tests dashboard authors. All rights reserved."?

mattl:
All rights reserved doesn't really fit with a 3-clause BSD program, because we're not reserving all rights.

mattl:
I'd go with Copyright 2014-2017 Web Platform Tests Dashboard Project. Licensed under the terms of <license name>. A URL to the license would be good too.

foolip:
My only nit with that is that I try to never write Web Platform Tests but always web-platform-tests or an abbreviation, would "WPT Dashboard Project" be OK?

rwaldron:
@mattl @foolip status?

mattl:
I think that would be fine @foolip

foolip:
OK, now updated!

mattl:
Merged.

Issue #40, opened 2017-07-14
Make test runner test against the first SHA of the day


jeffcarp:
Instead of having a scheduled job update the SHA all test runners should be testing against, we can just have all runners test against the first SHA committed that day.

foolip:
I don't think this can be made robust. Whatever the condition for testing commits for "is today" is and whatever order one takes at merge commits, later (merge) commits could make it such that another commit is the first that matches the "is today" condition. This is just a problem with Git, and that one can't tell what the master branch has pointed to in the past.

What is the current state of this, is the job that picks a commit still running?

jgraham:
Assuming each commit comes from a PR, you could find the first PR to merge today (whenever that is), and use the merge_commit_sha property to get the corresponding HEAD of master. It does add a dependency on the GitHub API though.

foolip:
Yeah, something around PR status should be robust against later changes in Git history. `curl https://api.github.com/repos/w3c/web-platform-tests/pulls/7233` has `"merged_at": "2017-09-04T07:19:20Z",63763763ad8fc064ad03ec1fa"` and as long as a PR isn't opened and merged again I guess that would never change.

I think one would have to inspect all 7000+ PRs, however, since any of the PRs could be the earliest one merged today, and agreeing on (say) 100 PRs but checking at different times could lead to different results.

Overall, I think it'd be less work for everyone to pick a commit per day and publish it somewhere. It would be fairly simple to set up a cron job that adds a new tag in the repo itself once per day, would that be better or worse than having the dashboard publish it?

jgraham:
I agree that your method is simpler. But something like:
```
curl 'https://api.github.com/search/issues?q=type:pr+repo:w3c/web-platform-tests+is:merged+merged:>=2017-09-04'
```
Can give you PRs that were merged today.

foolip:
Ah, I guess as long as the number of PRs in a day never exceeds 100, no pagination code would be needed, and that'd be robust.

I think this choice is ultimately up to @bobholt and what @thejohnjansen would prefer to work with.

gsnedders:
A potentially stupid but workable solution would be to simply store the "current test run commit" somewhere, and if the commit date of that commit is greater than 24 hours ago then change it to the current tip of master. No faffing around trying to figure out the "first commit of the day" or anything—just set it to be "one commit per day" with it chosen pretty arbitrarily.

jeffcarp:
> simply store the "current test run commit" somewhere

[...]

PR #245, opened 2017-11-14
Find a complete SHA when ?complete=true is present for /api/runs

lukebjerring:
Fixes https://github.com/w3c/wptdashboard/issues/138 (fix++)

Looks like the issue is that we can have duplicated data. Tested locally with the same data visible in https://wpt.fyi/test-runs

/cc @mattl


lukebjerring:
Pause for thought, though, the most recent complete run as of now (2017-11-14) is https://wpt.fyi/?sha=2c04684030 (2017-10-19). Nearly a month old. We should probably resolve https://github.com/w3c/wptdashboard/issues/40 before submitting this 'fix'.

jeffcarp:
Re: this approach - I was thinking would it be preferable in terms of performance to set up a cron job that runs every 10 minutes and calls `getLastCompleteRunSHA` and stores that result in a Datastore entity with a fixed key, like `LatestRun`? That sounds faster but the difference may be negligible. This approach sgtm, if we see an increase in latency we can follow it up with that optimization.

test_handler.go
```diff
@@ -86,3 +100,37 @@ func ParseSHAParam(r *http.Request) (runSHA string, err error) {
...
+
+	// Map is sha -> browser -> seen yet?  - this prevents over-counting dupes.
+	runSHAs := make(map[string]map[string]bool)
+	for _, browser := range browserNames {
+		it := baseQuery.Filter("BrowserName = ", browser).Run(ctx)
+		for {
+			var testRun TestRun
+			_, err := it.Next(&testRun)
+			if err == datastore.Done {
+				break
```
jeffcarp: If I'm reading this correctly, it breaks once it finds a match across all 4 browsers, correct? So in practice it won't be querying 400 entities, it will end up being closer to 10-20?

lukebjerring: No - I can (potentially) tweak it to have 4 iterators, and one-by-one them in parallel, but currently it fully iterates 3 browsers, and partially iterates the 4th (falling out a little early)

lukebjerring:
Your thoughts around making this a general behaviour with a URI param make sense. I'll update the PR to only do this behaviour when a query-param is present (+ move it to /api/runs instead)

PR #246, opened 2017-11-14
Remove Google disclaimer

mattl:
As this project is under the W3C umbrella now, we can remove this Google-specific disclaimer


mattl:
Let me do that again with just the file we need

PR #250, opened 2017-11-15
Migrate /test-runs POST to /api/runs POST

lukebjerring:
Also improves some handler naming (too general for methods across the namespace) and fixes a bug with the populate_dev_data script (which is being deleted in https://github.com/w3c/wptdashboard/pull/248 anyway).

### DIFF
--- a/api_handlers.go
+++ b/api_handlers.go
@@ -16,8 +16,10 @@ package wptdashboard
 
 import (
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
 	"net/url"
+	"time"
 
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
@@ -66,13 +68,23 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write(testRunsBytes)
 }
 
-// apiTestRunHandler is responsible for emitting the test-run JSON a specific run,
+func apiTestRunHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "POST" {
+		apiTestRunPostHandler(w, r)
+	} else if r.Method == "GET" {
+		apiTestRunGetHandler(w, r)
+	} else {
+		http.Error(w, "This endpoint only supports GET and POST.", http.StatusMethodNotAllowed)
+	}
+}
+
+// apiTestRunGetHandler is responsible for emitting the test-run JSON a specific run,
 // identified by a named browser (platform) at a given SHA.
 //
 // URL Params:
 //     sha: SHA[0:10] of the repo when the test was executed (or 'latest')
 //     browser: Browser for the run (e.g. 'chrome', 'safari-10')
-func apiTestRunHandler(w http.ResponseWriter, r *http.Request) {
+func apiTestRunGetHandler(w http.ResponseWriter, r *http.Request) {
 	runSHA, err := ParseSHAParam(r)
 	if err != nil {
 		http.Error(w, "Invalid query params", http.StatusBadRequest)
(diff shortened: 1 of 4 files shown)


lukebjerring:
Note that this PR leaves the existing endpoint functional, as migrating the scripts that utilize it is best done before deleting the old endpoint.

rwaldron:
@lukebjerring quick rebase?

populate_dev_data_handler.go
```diff
@@ -86,11 +86,14 @@ func populateDevData(w http.ResponseWriter, r *http.Request) {
 		"safari",
 	} {
 		// TODO(lukebjerring): Move wpt.fyi base URL to constant.
-		jsonURL := "https://wpt.fyi/json?platform=" + browserName
+		jsonURL := "https://wpt.fyi/results?platform=" + browserName
```
jeffcarp: Is this diff related to this PR?

populate_dev_data_handler.go
```diff
@@ -86,11 +86,14 @@ func populateDevData(w http.ResponseWriter, r *http.Request) {
...
 		resp, err := client.Head(jsonURL)
 
-		if urlError, ok := err.(*url.Error); !ok || urlError.Err != errUseLastResponse {
-			fmt.Fprintf(w, "Failed to fetch latest run for %s: %s\n", browserName, urlError.Error())
+		if err == nil {
+			fmt.Fprintf(w, "Got unexpected non-redirect status %d for %s\n", resp.StatusCode, jsonURL)
+			continue
+		} else if urlError, ok := err.(*url.Error); !ok || urlError.Err != errUseLastResponse {
```
jeffcarp: The `err.(*url.Error)` construct scrambles my brain every time

api_handlers.go
```diff
@@ -120,6 +132,56 @@ func apiTestRunHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	if suppliedSecret != token.Secret {
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	}
+
+	var body []byte
+	if body, err = ioutil.ReadAll(r.Body); err != nil {
```
jeffcarp: I think you need to add `defer r.Body.Close()` somewhere in this function in order to not leak sockets.

test_run_handler.go
```diff
@@ -16,71 +16,26 @@ package wptdashboard
...
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
 )
 
-func testRunHandler(w http.ResponseWriter, r *http.Request) {
+// testRunsHandler handles GET/POST requests to /test-runs
+func testRunsHandler(w http.ResponseWriter, r *http.Request) {
 	if r.Method == "POST" {
-		handlePost(w, r)
+		// /test-runs is the legacy POST endpoint, migrated to /api/runs and left to avoid breakages.
```
jeffcarp: Can you make a bug and add `TODO(#n)` where `n` is the bug #?

api_handlers.go
lukebjerring: Done

populate_dev_data_handler.go
lukebjerring: Not really; I'll make this tweak separately.

populate_dev_data_handler.go
lukebjerring: Yeah casting isn't so intuitive written this way.

test_run_handler.go
lukebjerring: Done.

api_handlers.go
jeffcarp: Ah! TIL :+1:

PR #260, opened 2017-11-17
Add protos for Browser, TestRun and Token

lukebjerring:
Currently unused by the Go code (to keep the PR small!)


.gitignore
```diff
@@ -4,3 +4,4 @@ run/running.ini
 .DS_Store
 run/protos/*_pb2.py
 bq-schema
+generated
```
mdittmer: nit: can you config your editor to include a newline at the end of files?

protos/test_run.proto
```diff
@@ -0,0 +1,35 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package w3c.wptdashboard;
```
mdittmer: Should we bother with the `w3c` prefix? When I first put that in strawman.proto I was under the delusion that we could easily package-path our python and go code accordingly without hassle (which we can't, not without hassle).

mdittmer: ... if you're wondering, the package path matters (at the very least) because it determines the directory structure where BigQuery schema go (and we'll eventually have scripts that need to ingest them and setup/update schema using the `bq` tool).

.gitignore
lukebjerring: Done.

protos/test_run.proto
lukebjerring: Removed.

lukebjerring:
Will add tests in a follow-up PR.

PR #261, opened 2017-11-17
Two Python dep changes

mdittmer:
- Make sure wheel is installed before attempting to install other deps
- Add google-cloud-bigquery as a dep in preparation for pushing data to BigQuery


Makefile
```diff
@@ -67,6 +67,7 @@ py_proto: $(PROTOS)
 		--python_out=$(PB_PY_OUT_DIR) $(PROTOS)
 
 py_deps: $(find . -type f | grep '\.py$' | grep -v '\_pb.py$')
+	pip install wheel===0.30.0
```
lukebjerring: Maybe comment why it's separated.

mdittmer:
@Hexcles PTAL: https://travis-ci.org/w3c/wptdashboard/jobs/303680532#L3084

This arose from using `pip freeze > requirements.txt`.

Hexcles:
I just experimented in a local virtualenv. Two notes:

* wheel is indeed an optional dependency, but only if you have gcc... Without wheel, pip will fall back to compile from source tarball which apparently would fail in a minimal docker image.
* bzr is definitely not depended by anything in this project. I suspect the environment you froze was not originally clean. Try to dump out `pip freeze` before installing py deps.

Hexcles:
Cool! Glad that those pre-installed packages turned out easy to be removed and didn't become a cascading hell.

I was about to say that there was really no "perfect" solution. The two suboptimal solutions I could think of were: 1. removing these pre-installed packages to make the system as clean as possible, which is not always feasible; 2. switching back to using virtualenv and installing all dependencies there, but that means we need to use `wpt --venv` (to stop it from creating messy nested virtualenv), which is a much less used code path.

PR #263, opened 2017-11-17
Add gofmt check to go_lint rule

lukebjerring:
/cc @jeffcarp

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ py_test: py_proto py_deps
 
 go_lint: go_deps
 	golint -set_exit_status
+    diff -u <(echo -n) <(gofmt -d ./)
 
 go_test: go_deps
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,6 +52,8 @@ py_test: py_proto py_deps
 
 go_lint: go_deps
 	golint -set_exit_status
+	# Print differences between current/gofmt'd output, check empty.
```
mdittmer: :+1: :100:

Makefile
```diff
@@ -52,6 +52,8 @@ py_test: py_proto py_deps
 
 go_lint: go_deps
 	golint -set_exit_status
```
mdittmer: Is there a way to pass golint something that will do this for us?

Issue #251, opened 2017-11-15
Migrate POST request for /test-runs to /api/runs


lukebjerring:
REST-esque methods and JSON endpoints live under /api/.

lukebjerring:
Started by https://github.com/w3c/wptdashboard/pull/250

PR #271, opened 2017-11-20
Migrate from POST /test-runs to POST /api/run

lukebjerring:
Contributes to resolving https://github.com/w3c/wptdashboard/issues/251


lukebjerring:
The POST handler (optionally), but the /test-runs is an endpoint for serving the HTML.

PR #280, opened 2017-11-21
Add support for specifying which browsers in /api/runs

lukebjerring:
Params handler works for both

    ?browser=chrome&browser=safari

and 

    ?browsers=chrome,safari


params.go
```diff
@@ -0,0 +1,89 @@
...
+func ParseSHAParam(r *http.Request) (runSHA string, err error) {
+	// Get the SHA for the run being loaded (the first part of the path.)
+	runSHA = "latest"
+	params, err := url.ParseQuery(r.URL.RawQuery)
+	if err != nil {
+		return runSHA, err
+	}
+
+	runParam := params.Get("sha")
+	regex := regexp.MustCompile("[0-9a-fA-F]{10}")
```
mdittmer: This could be a pre-compiled constant outside the func.

params.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	browser = r.URL.Query().Get("browser")
+	// Check that it's a browser name we recognize.
+	for _, name := range browserNames {
+		if name == browser {
+			return name, nil
+		}
+	}
+	return "", nil
```
mdittmer: ditto (logging)

params.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return runSHA, err
+	}
+
+	runParam := params.Get("sha")
+	regex := regexp.MustCompile("[0-9a-fA-F]{10}")
+	if regex.MatchString(runParam) {
+		runSHA = runParam
+	}
+	return runSHA, err
```
mdittmer: Can we log here? No error will be reported but we're falling back on something other than what was input to the function.

params.go
```diff
@@ -0,0 +1,89 @@
...
+	if browsersParam := r.URL.Query().Get("browsers"); browsersParam != "" {
+		browsers = append(browsers, strings.Split(browsersParam, ",")...)
+	}
+	// If no params found, return the default.
+	var browserNames []string
+	if browserNames, err = GetBrowserNames(); err != nil {
+		return nil, err
+	}
+	if len(browsers) == 0 {
+		return browserNames, nil
```
mdittmer: ditto (logging)

params.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	// Otherwise filter to valid browser names.
+	for i := 0; i < len(browsers); {
+		if !IsBrowserName(browsers[i]) {
+			// 'Remove' browser by switching to end and cropping.
```
mdittmer: If you say so :smiley: @jeffcarp can you please golang-review this bit?

mdittmer: Log this case?

params.go
```diff
@@ -0,0 +1,89 @@
...
+		if name == browser {
+			return name, nil
+		}
+	}
+	return "", nil
+}
+
+// ParseBrowsersParam parses the 'browsers' parameter, split on commas.
+// If empty, it also checks for the (repeatable) 'browser' params, before falling back to the default set of
+// browsers.
```
mdittmer: Mention returns sorted list?

params_test.go
```diff
@@ -0,0 +1,142 @@
...
+func TestParseSHAParam(t *testing.T) {
+	r := httptest.NewRequest("GET", "http://wpt.fyi/", nil)
+	runSHA, err := ParseSHAParam(r)
+	assert.Nil(t, err)
+	assert.Equal(t, "latest", runSHA)
+}
+
+func TestParseSHAParam_2(t *testing.T) {
+	sha := "0123456789"
+	r := httptest.NewRequest("GET", "http://wpt.fyi/?sha="+sha, nil)
```
mdittmer: nit: ` + sha`

api_handlers.go
```diff
@@ -52,7 +51,7 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var browserNames []string
-	if browserNames, err = GetBrowserNames(); err != nil {
+	if browserNames, err = ParseBrowsersParam(r); err != nil {
```
mdittmer: @jeffcarp is this special case of multiple-statements-on-one-line-with-error-checking a standard golang idiom?

lukebjerring: Yep, they're called 'short statements' - https://tour.golang.org/flowcontrol/6

params_test.go
lukebjerring: These files are gofmt'd, that's the way they like it :confused:

params.go
lukebjerring: It's user-input, so maybe we should just throw an error. Either way - refactor or plumbing needed, so will follow-up. https://github.com/w3c/wptdashboard/issues/283

PR #282, opened 2017-11-21
Migrate jenkins.py 'runner script' to Runner class

mdittmer:


run/runner.py
```diff
@@ -0,0 +1,387 @@
...
+            self.write_gzip_json(filepath, result)
+            print('Wrote file %s' % filepath)
+
+    def upload_results(self):
+        command = ['gsutil', '-m', '-h', 'Content-Encoding:gzip',
+                   'rsync', '-r', self.sha, 'gs://wptd/%s' % self.sha]
+        return_code = subprocess.check_call(command, cwd=self.output_path)
+        assert return_code == 0, 'gsutil rsync return code was not 0!'
+
+    def upload_run(self):
```
JKereliuk: I like that these commands are now broken up like this. Easier to read

run/runner.py
```diff
@@ -0,0 +1,387 @@
...
+        local_log_filepath=None,
+        local_report_filepath=None,
+        summary_path=None,
+        local_summary_gz_filepath=None,
+        gs_results_filepath_base=None,
+        gs_http_results_url=None,
+        summary_filename=None,
+        summary_http_url=None,
+    ):
+        self.metadata_url = metadata_url
```
JKereliuk: nit: I think we should have a plan to document what all of these are, maybe just a short sentence but would be nice to have in the future. Most of these are self explanatory by the name, but also requires a little context.

run/jenkins_test.py
```diff
@@ -14,14 +14,13 @@
```
JKereliuk: Can we move this to a test folder?

mdittmer: Pardon my Python ignorance: Would that require us to use relative imports that include `..` or something? If someone can help me sort out how the test code refers to the code it's testing I'm happy to make the change. @jeffcarp is there a reason we have been using _test suffix instead of a test dir?

run/runner.py
mdittmer: I agree, we should have a follow-up PR for that. I'm not convinced this interface is "mature" yet. Some config to control push-to-BigQuery (vs. / alongside push-to-GCS) will be added soon, for example.

run/jenkins_test.py
JKereliuk: Whoops, I misread when I looked at the code before. I didn't see we already had ```_test``` suffixed files.
In that case, I think we should merge this PR with the file organization as is here and change browsers_test and run_test in another PR afterwards. I will open an issue.

PR #285, opened 2017-11-21
Git hooks

mdittmer:
Add pre-push git hook that will run `make lint` inside a docker container. To enable, symlink `.git/hooks -> git/hooks` or just `.git/hooks/pre-push -> git/hooks/pre-push`.


git/hooks/pre-push
```diff
@@ -0,0 +1,48 @@
...
+source "${REPO_DIR}/util/logging.sh"
+source "${REPO_DIR}/util/path.sh"
+WPTD_PATH=${WPTD_PATH:-$(absdir "${REPO_DIR}")}
+
+docker inspect wptd-dev-instance > /dev/null 2>&1
+INSPECT_STATUS="${?}"
+
+DOCKER_INSTANCE_PID=""
+if [ "${INSPECT_STATUS}" != "0" ]; then
+  docker run -t -d --entrypoint /bin/bash \
```
lukebjerring: Do the logs from running an instance (if not already running) show in the console? If not, worth logging "Starting docker instance" or similar.

git/hooks/pre-push
```diff
@@ -0,0 +1,48 @@
...
+      -u $(id -u $USER):$(id -g $USER) \
+      --name wptd-dev-instance wptd-dev &
+  DOCKER_INSTANCE_PID="${!}"
+fi
+
+docker exec -t -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make lint
+LINT_STATUS="${?}"
+FINAL_STATUS="${LINT_STATUS}"
+if [ "${LINT_STATUS}" != "0" ]; then
+  warn "Lint failed. Push anyway (y/N)?"
```
lukebjerring: If you take a look at the sample, they pass in a flag, e.g. --ignore-lint-errors.
However, I like this better.

mdittmer: Let's keep it as-is. Follow-up PR if sticking around to press `y` gets annoying.

PR #287, opened 2017-11-22
Fix go_lint make target by tweaking go package and Makefile

mdittmer:


Makefile
```diff
@@ -55,7 +57,7 @@ py_test: py_proto py_deps
 go_lint: go_deps
 	golint -set_exit_status
 	# Print differences between current/gofmt'd output, check empty.
-	! gofmt -d ./... 2>&1 | read
+	! gofmt -d . 2>&1 | read
```
lukebjerring: These 2 changes feel unrelated; go_package just affects the package statement in the go file itself, while this is saying 'top-level go files only'?

mdittmer: > this is saying 'top-level go files only'

In practice that doesn't appear to be the case. Here's what I see:

```bash
# without Makefile change:

docker exec -t -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make go_lint
# Output:
# /bin/sh: 1: read: arg count

# Make sure `make` uses /bin/bash

docker exec -t -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make go_lint
# Output:
# Makefile:58: recipe for target 'go_lint' failed

# Drop pipe-to-read to see error output

docker exec -t -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make go_lint
# Output:
# stat ./...: no such file or directory

# Switch ./... to .; deliberately malformat generated/browser.pb.go
# This should confirm that subdirectories are hit by "." and "./..." shouldn't be necessary

docker exec -t -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make go_lint
# Output:
# diff generated/browser.pb.go gofmt/generated/browser.pb.go
# --- /tmp/gofmt218530372 2017-11-22 16:06:10.012714713 +0000
# +++ /tmp/gofmt911124691 2017-11-22 16:06:10.012714713 +0000
# ...

# Re-add pipe-to-read; lint fails until generated/browser.pb.go is fixed.
```

PR #291, opened 2017-11-22
Add IsBrowserName util method, and test it.

lukebjerring:
This actually highlighted a bug where a function was re-using a global variable unintentionally (see getLastCompleteRunSHA).

### DIFF
--- a/api_handlers.go
+++ b/api_handlers.go
@@ -215,7 +215,8 @@ func getBrowserParam(r *http.Request) (browser string, err error) {
 	return "", nil
 }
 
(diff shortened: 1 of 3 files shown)


api_handlers.go
```diff
@@ -209,23 +209,14 @@ func getBrowserParam(r *http.Request) (browser string, err error) {
...
-		if name == browser {
-			return name, nil
-		}
+	if browser = params.Get("browser"); IsBrowserName(browser) {
+		return browser, nil
 	}
 	return "", nil
 }
 
 // getLastCompleteRunSHA returns the SHA[0:10] for the most recent run that complete for all of the given browser names.
```
mdittmer: nit: update docs

PR #294, opened 2017-11-23
Use first sha for the day

lukebjerring:
Fixes https://github.com/w3c/wptdashboard/issues/40

Example output:

    run/shas.py --wpt-path ../web-platform-tests/
    INFO:root:HEAD SHA is 6e1df92d1080ef918244785e259bc3f4507cd88a
    INFO:root:First SHA for 2017-11-23 is 2a412bfbec794ce4debf92c73f6b85bf146f74ba


run/run.py
```diff
@@ -113,10 +119,18 @@ def main(platform_id, platform, args, config):
 
     patch_wpt(config, platform)
 
-    # TODO(#40): modify this to test against the first SHA of the day
-    CURRENT_WPT_SHA = get_current_wpt_sha(config)
+    wpt_path = config['wpt_path']
+    sha_finder = shas.SHAFinder(logger)
+    CURRENT_WPT_SHA = (sha_finder.get_todays_sha(wpt_path)
```
foolip: Rename this to something more accurate, like just WPT_SHA, or maybe CHOSEN_WPT_SHA?

run/shas.py
```diff
@@ -0,0 +1,108 @@
...
+        Gets the first SHA for the current date of the git repo in the given
+        path.
+        """
+        today = date.today()
+        tomorrow = today + timedelta(days=1)
+        command = [
+            'git',
+            'log',
+            '--format=%H',
+            '--after="%s 00:00:00"' % today.isoformat(),
```
foolip: Use "T00:00Z" so that it's UTC time regardless of what the system time is.

run/shas.py
```diff
@@ -0,0 +1,108 @@
...
+            '--before="%s 00:00:00"' % tomorrow.isoformat(),
+            '--date-order',
```
foolip: `--date-order` will make this less stable than it could be I think. Just dropping that should result first commit in tree order after the cutoff being used, even if there later appears a commit with an earlier timestamp. Might want to verify with tests though :)

run/shas.py
```diff
@@ -0,0 +1,108 @@
...
+        abspath = os.path.abspath(path)
+        logger.debug('Using dir ' + abspath)
+        logger.debug('Executing ' + ' '.join(command))
+        output = subprocess.check_output(command, cwd=abspath)
+        sha = output.decode('UTF-8').strip()
+        assert len(sha) == 40, 'Invalid SHA: "%s"' % sha
+        return sha
+
+
+if __name__ == '__main__':
```
foolip: This is never called in prod or tests, right? A nice-to-have?

run/run.py
lukebjerring: Done.

run/shas.py
```diff
@@ -0,0 +1,108 @@
...
+
+from datetime import date, timedelta
+
+"""
+shas.py defines methods for determining which SHA to run against on a given
+date.
+"""
+
+
+class SHAFinder(object):
```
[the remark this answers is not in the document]
lukebjerring: It'd be kinda pointless.
Unit test coverage would be:
- do I call a subprocess with the flags that the unit test says I'm supposed to?

Integration test is useful (hence ability to run shas.py).

Added a test against wptd (itself as the repo) on a fixed date.

run/shas.py
lukebjerring: Sure - it'll generally be run from the same machine/cluster, except in retroactive runs.. but Done.

run/shas.py
lukebjerring: Test still passes.. so sure :)

run/shas.py
lukebjerring: Yeah, I wrote/used for local integration testing.

run/shas.py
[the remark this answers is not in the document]
foolip: A unit test would be pointless yes, running against real repos is what I had in mind.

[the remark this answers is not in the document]
foolip: Should that test be part of this PR?

run/shas_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class TestSHAFinder(unittest.TestCase):
+
+    def test_nov_21st(self):
+        logger = mock.Mock(logging.Logger)
+
+        sha_finder = shas.SHAFinder(logger, date(2017, 11, 21))
+
+        self.assertEqual('46060eb2c33de6101bc6930bf5e34f794aa9f996',
```
foolip: Trying to find another commit to test with, I think I'm wrong about `--date-order`. Removing it doesn't make any difference in all commits since 2017-01-01, but adding `--topo-order` on the other hand does.

Just one test without the edge case is fine, but I guess if you wanted to find such a case in history, the easiest way would be to loop over dates here until you find one where adding `--topo-order` gives a different result, and picking which is the best behavior.

run/shas.py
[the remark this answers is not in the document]
lukebjerring: Was missing, added.

gsnedders:
This seems quite bad; there's no guarantee that the commit that gets run was ever the top of the master branch. I'd say we only want to really trust commits that have at some point been the top of the master branch (everything else won't have run through CI and we no idea if it's good or not).

lukebjerring:
I think the likelyhood of a bad unsquashed commit, is low, no?

We do have a general objective of whitelisting SHAs worth testing (once we get > 1 per day), which will cover this potential problem.

PR #298, opened 2017-11-24
Add support for max-count param in /api/runs

lukebjerring:


foolip:
A description in the first commit / description would be very nice, both for context when reviewing and when finding this in git blame later.

api_handlers.go
```diff
@@ -57,7 +58,15 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var testRuns []TestRun
-	baseQuery := datastore.NewQuery("TestRun").Order("-CreatedAt").Limit(1)
+	var limit int
+	if limit, err = ParseMaxCountParam(r); err != nil {
+		http.Error(w, "Invalid 'max-count' param: "+err.Error(), http.StatusBadRequest)
```
foolip: Aside: I don't know Go, but am wondering if runtime error exist, and if so how easy/hard it is to write error handling code that's untested and will actually blow up in a different way than intended if run. (Reminds me of a C++ code base I worked on with a lot of out-of-memory handling that was untested.)

params.go
```diff
@@ -85,3 +95,20 @@ func ParseBrowsersParam(r *http.Request) (browsers []string, err error) {
 	sort.Strings(browsers)
 	return browsers, nil
 }
+
+// ParseMaxCountParam parses the 'max-count' parameter as an integer.
+func ParseMaxCountParam(r *http.Request) (count int, err error) {
+	count = MaxCountDefaultValue
+	if browsersParam := r.URL.Query().Get("max-count"); browsersParam != "" {
```
foolip: Should this by maxCountParam or countParam instead of browsersParam?

params_test.go
```diff
@@ -140,3 +140,31 @@ func TestParseBrowsersParam_MultiBrowserParam_AllInvalid(t *testing.T) {
...
+
+func TestParseMaxCountParam_Missing(t *testing.T) {
+	r := httptest.NewRequest("GET", "http://wpt.fyi/", nil)
+	count, err := ParseMaxCountParam(r)
+	assert.Nil(t, err)
+	assert.Equal(t, MaxCountDefaultValue, count)
+}
+
+func TestParseMaxCountParam_TooSmall(t *testing.T) {
+	r := httptest.NewRequest("GET", "http://wpt.fyi/?max-count=0", nil)
```
foolip: Also try max-count=-1 or max-count=1e8 to see that the "not a non-negative integer" error handling is right?

lukebjerring:
Commit history cleaned up, PTAL

util.go
```diff
@@ -47,6 +47,11 @@ func GetBrowsers() (map[string]Browser, error) {
 // GetBrowserNames returns an alphabetically-ordered array of the names
 // of the browsers returned by GetBrowsers.
 func GetBrowserNames() ([]string, error) {
+	if browserNamesAlphabetical == nil {
```
foolip: Is this change accidentally in this PR, or was it a necessary adjustment?

api_handlers.go
lukebjerring: Some reading for your leisure time: https://blog.golang.org/defer-panic-and-recover

params.go
lukebjerring: Fixed (CopyPasta)

params_test.go
lukebjerring: Done.

PR #300, opened 2017-11-24
Add 'path' filter to /api/diff

lukebjerring:
Progresses https://github.com/w3c/wptdashboard/issues/301

Running @ https://20171124t141915-dot-wptdashboard.appspot.com

Example:
https://20171124t141915-dot-wptdashboard.appspot.com/api/diff?before=safari@fc2e57a502&after=safari@e0a58d4cac&path=/css/vendor-imports/


foolip:
This is also ultimately for https://github.com/w3c/web-platform-tests/issues/7475, right? Is there a tracking issue in this repo that outlines the bits and pieces needed that PRs like this one could link to?

params.go
```diff
@@ -21,6 +21,8 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+
+	"github.com/deckarep/golang-set"
```
foolip: Aside: I see that this set has a contains method. Maybe use this more when array membership requires a loop :)

params.go
```diff
@@ -85,3 +87,24 @@ func ParseBrowsersParam(r *http.Request) (browsers []string, err error) {
 	sort.Strings(browsers)
 	return browsers, nil
 }
+
+// ParsePathsParam returns a set list of test paths to include, or nil if no filter is provided (and all tests should be
+// included). It parses the 'paths' parameter, split on commas, and also checks for the (repeatable) 'path' params.
```
foolip: Supporting both is harmless, but is anything made simpler on the client side by this? Seems like more things to wonder about, like "can I provide paths multiple times too"? (No.)

params.go
```diff
@@ -85,3 +87,24 @@ func ParseBrowsersParam(r *http.Request) (browsers []string, err error) {
...
+		return nil
+	}
+
+	paths = mapset.NewSet()
+	for _, path := range pathParams {
+		paths.Add(path)
+	}
+	if browsersParam := pathsParam; browsersParam != "" {
+		for _, path := range strings.Split(browsersParam, ",") {
+			paths.Add(path)
```
foolip: Should empty strings be added here, or above? Tests for the various ways to get empty strings here and above would be good.

params.go
```diff
@@ -85,3 +87,24 @@ func ParseBrowsersParam(r *http.Request) (browsers []string, err error) {
...
+	pathsParam := r.URL.Query().Get("paths")
+	if len(pathParams) == 0 && pathsParam == "" {
+		return nil
+	}
+
+	paths = mapset.NewSet()
+	for _, path := range pathParams {
+		paths.Add(path)
+	}
+	if browsersParam := pathsParam; browsersParam != "" {
```
foolip: browsersParam here looks curious, should it just turn into `if pathsParam != "" {`? If so then no guard at all should be necessary, just let the empty string be split.

params_test.go
```diff
@@ -140,3 +140,27 @@ func TestParseBrowsersParam_MultiBrowserParam_AllInvalid(t *testing.T) {
...
+	r := httptest.NewRequest("GET", "http://wpt.fyi/api/diff?paths=/css,/css", nil)
+	paths := ParsePathsParam(r)
+	assert.Len(t, paths.ToSlice(), 1)
+}
+
+func TestParsePathsParam_PathsAndPath_Duplicate(t *testing.T) {
+	r := httptest.NewRequest("GET", "http://wpt.fyi/api/diff?paths=/css&path=/css", nil)
+	paths := ParsePathsParam(r)
+	assert.Len(t, paths.ToSlice(), 1)
+}
```
foolip: Additional tests:
* `?path=` (empty string)
* `?paths=/dom,,` (empty strings)
* `?paths=/css&paths=/dom,/html` (first, last, or both used?)

foolip: Um, it's possible I'm thinking too much like I'm reviewing browser code where the edge cases have to be deliberately picked. If you think that's not helpful in this kind of code, please say so.

run_diff.go
```diff
@@ -104,6 +105,7 @@ func fetchRunResultsJSON(ctx context.Context, r *http.Request, run TestRun) (res
 	if strings.Index(url, "/") == 0 {
 		reqURL := *r.URL
 		reqURL.Path = url
+		panic(reqURL)
```
foolip: Is this leftover debugging?

run_diff.go
```diff
@@ -147,10 +153,26 @@ func diffResults(before map[string][]int, after map[string][]int) map[string][]i
 		}
 	}
 	for test, resultsAfter := range after {
+		if paths != nil && !anyPathMatches(paths, test) {
```
foolip: I think the reason for the nil check here and not above is intentional, but that would be clearer if both were handled outside of anyPathMatches, or with a comment.

run_diff.go
```diff
@@ -147,10 +153,26 @@ func diffResults(before map[string][]int, after map[string][]int) map[string][]i
...
 			// Missing? Then N / N tests are 'different'
 			diff[test] = []int{resultsAfter[1], resultsAfter[1]}
 		}
 	}
 	return diff
 }
+
+func anyPathMatches(paths mapset.Set, testPath string) bool {
+	if paths == nil {
+		return true
```
foolip: If nil is a way to represent an empty set, then this is the opposite of what I'd expect. Maybe assert that it's not null and handle it outside?

util.go
```diff
@@ -47,6 +47,11 @@ func GetBrowsers() (map[string]Browser, error) {
 // GetBrowserNames returns an alphabetically-ordered array of the names
 // of the browsers returned by GetBrowsers.
 func GetBrowserNames() ([]string, error) {
+	if browserNamesAlphabetical == nil {
```
foolip: Here it is again, what's going on? :)

run_diff_test.go
```diff
@@ -69,17 +71,51 @@ func TestDiffResults_Removed(t *testing.T) {
 	assertDelta(t, []int{1, 2}, []int{1, 1}, []int{1, 2})
 }
 
+func TestDiffResults_Filtered(t *testing.T) {
```
foolip: I didn't review this very carefully, hope it's correct :)

params.go
lukebjerring: That's the result of copy + paste + intellij refactor out into a variable. Fixed.

params.go
lukebjerring: Empty string is non-sensical. Removed below.

run_diff.go
lukebjerring: Sure is.

run_diff.go
lukebjerring: nil is 'no filter set', which is not the same as an empty set.

run_diff.go
lukebjerring: Not deliberate. Fixed.

lukebjerring:
Merged master, and adapted to play nicely with the ?filter=[A][C][D] behaviour from https://github.com/w3c/wptdashboard/pull/303. 

PTAL

foolip:
I started this review, but adding @mattl and @rwaldron in keeping with the email I just sent out.

rwaldron:
@lukebjerring for future commit conflict resolution, please use "git pull --rebase upstream master" (assuming you have a remote called "upstream" that points to this repo). Thanks!

lukebjerring:
OK - I have been leaving the merge as a separate commit in my PRs to make it obvious for the review / comment system, and later squashing on merge to master (in github).

Is that not a flow that you agree with? Rebasing has a confusing impact on existing review comments.

PR #302, opened 2017-11-27
Migrate /test-runs and / to use /api/runs

lukebjerring:
For the index/homepage, this changes 4 /api/run fetch requests to a single /api/runs request.

For /test-runs, it defers the data-store fetch to the request from the client, greatly simplifying the Go code, with minimal effect to latency.

Running here:
https://20171127t105621-dot-wptdashboard.appspot.com

Note that this also allows for a bigger fetch of test-runs, which is related to https://github.com/w3c/wptdashboard/issues/73 - e.g.:
https://20171127t105621-dot-wptdashboard.appspot.com/test-runs?max-count=500

### DIFF
--- a/components/wpt-runs.html
+++ b/components/wpt-runs.html
@@ -136,7 +136,7 @@ limitations under the License.
       }
 
       _computeDateTooltip(date) {
-        return date.toDateString()
+        return date && date.toDateString()
       }
 
       async connectedCallback () {
--- a/components/runs.html
+++ b/components/runs.html
@@ -51,7 +51,13 @@ limitations under the License.
                 return response.json()
               })
             )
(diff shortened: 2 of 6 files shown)


components/wpt-runs.html
```diff
@@ -136,7 +136,7 @@
       }
 
       _computeDateTooltip (date) {
-        return date.toDateString()
+        return date && date.toDateString()
```
foolip: What's this for? I see nothing related to _computeDateTooltip and nothing related in the commit message.

lukebjerring: When the computation occurs on bad data, this just prevents an exception.

Issue #301, opened 2017-11-24
Surface API for diffing two runs


lukebjerring:
Ultimate goal is to be able to detect regressions, for the issue in web-platform-tests: https://github.com/w3c/web-platform-tests/issues/7475

TODO: Design Doc

lukebjerring:
This issue was moved to web-platform-tests/wpt.fyi#20

PR #303, opened 2017-11-27
Add support for filtering change types when diffing

lukebjerring:
Progresses https://github.com/w3c/wptdashboard/issues/301

Uses a git-inspired filtering charset param.

?filter=[A][C][D]
A = Added
C = Changed
D = Deleted

### DIFF
--- a/params.go
+++ b/params.go
@@ -112,3 +112,43 @@ func ParseMaxCountParam(r *http.Request) (count int, err error) {
 	}
 	return count, err
 }
+
+// DiffFilterParam represents the types of changed test paths to include.
+type DiffFilterParam struct {
+	// Added tests are present in the 'after' state of the diff, but not present
+	// in the 'before' state of the diff.
+	Added bool
+
+	// Deleted tests are present in the 'before' state of the diff, but not present
+	// in the 'after' state of the diff.
+	Deleted bool
+
+	// Changed tests are present in both the 'before' and 'after' states of the diff,
+	// but the number of passes, failures, or total tests has changed.
+	Changed bool
+}
+
+// ParseDiffFilterParam splits the filter param into the differences to include.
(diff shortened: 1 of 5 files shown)


run_diff.go
```diff
@@ -124,32 +124,43 @@ func fetchRunResultsJSON(ctx context.Context, r *http.Request, run TestRun) (res
...
-			countDiff := abs(resultsBefore[1] - resultsAfter[1])
-			if countDiff == 0 && passDiff == 0 {
-				continue
-			}
-			// Changed tests is at most the number of different outcomes,
-			// but newly introduced tests should still be counted (e.g. 0/2 => 0/5)
-			diff[test] = []int{
-				max(passDiff, countDiff),
-				max(resultsBefore[1], resultsAfter[1]),
+	if filter.Deleted || filter.Changed {
```
lukebjerring: This diff tool isn't great. FWIW, changes are just:
- Surround whole code-block in the if-statement above
- Add continue if-statements

params.go
```diff
@@ -112,3 +112,43 @@ func ParseMaxCountParam(r *http.Request) (count int, err error) {
...
+	param = DiffFilterParam{
+		true,
+		true,
+		true,
+	}
+	if filter := r.URL.Query().Get("filter"); filter != "" {
+		param = DiffFilterParam{}
+		for _, char := range filter {
+			switch char {
+			case 'A':
```
foolip: Comment around here as well that this is inspired by `git --diff-filter=`, which, BTW, I really like :) I don't think that rename detection is in any way a straightforward thing, or likely to ever be worth doing, but if we do, it should be 'R'.

run_diff.go
foolip: Thanks, very helpful.

params.go
lukebjerring: Noted, and Done.

PR #305, opened 2017-11-28
Revert "Dockerize WPT runs & add Jenkins k8s specs" and following changes

foolip:
https://github.com/w3c/wptdashboard/issues/273


foolip:
Everyone, before Thanksgiving @mattl said that his plan was to try reverting the Jenkins changes to go back to using plain VMs in order to fix the urgent issue https://github.com/w3c/wptdashboard/issues/273. However, Matt is out sick, so this is my attempt to do just that.

I wanted to revert 17eb3274f078b17f210fcc09f33990a60c46e71f, and reverted things that depended on it to get back to a state where Docker and Jenkins are not mentioned in the project. Most likely I reverted a bit too much, or this would leave the project in a broken state.

Please review carefully, and suggest other ways of getting wpt.fyi updated ASAP if this does not seem like the quickest way.

git/hooks/pre-push
```diff
@@ -1,49 +0,0 @@
...
-      -u $(id -u $USER):$(id -g $USER) \
-      --name wptd-dev-instance wptd-dev &
-  DOCKER_INSTANCE_PID="${!}"
-fi
-
-docker exec -t -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make lint
-LINT_STATUS="${?}"
-FINAL_STATUS="${LINT_STATUS}"
-if [ "${LINT_STATUS}" != "0" ]; then
-  warn "Lint failed. Push anyway (y/N)?"
```
foolip: @mdittmer is there a way to preserve this while not depending on Docker?

foolip:
@lukebjerring @mdittmer @JKereliuk, would doing this huge revert that we can't do staging setups like in https://github.com/w3c/wptdashboard/pull/228?

foolip:
https://github.com/w3c/wptdashboard/pull/307 should fix the Travis failures seen here.

rwaldron:
@foolip before we do this, I'd like to have some time to re-review all of the changes leading to this revert. Thanks!

foolip:
@rwaldron, thank you, I think that would be very wise :)

rwaldron:
Do we have a way to do a test build with this PR?

JKereliuk:
@rwaldron Are you looking for the [travis build](https://travis-ci.org/w3c/wptdashboard/builds/308482526)?

rwaldron:
@JKereliuk maybe, but I'm not sure—I'm trying to understand how/where this could be staged to confirm that it corrects the issues in #273. The switch to docker occurred before I was present, so I'm struggling with how I might help confirm that this will resolve (even if temporarily) the issues. When I look at travis, it shows that master is green as well, even though the test runner code is not functional on jenkins

foolip:
@mattl @rwaldron, please go ahead and close this if you don't think it's our best shot at stability.

mattl:
@rwaldron Let's chat about this tomorrow.

PR #308, opened 2017-11-28
Add Docker.dev contents to Docker.jenkins

rwaldron:
@foolip this might be wrong, I'm not sure, but maybe worth a try?


rwaldron:
Ugh nevermind, I had an out of date master. /me facepalm

PR #309, opened 2017-11-28
Check for local changes in git before pushing

lukebjerring:
Fairly easy to make tweaks locally that, for example, fix lint, and then run push without committing them. This just checks that a git diff has no output (and can be bypassed with a Y confirmation)

### DIFF
--- a/git/hooks/pre-push
+++ b/git/hooks/pre-push
@@ -12,7 +12,7 @@ WPTD_PATH=${WPTD_PATH:-$(absdir "${REPO_DIR}")}
 DIFF_STATUS="${?}"
 if [ "${DIFF_STATUS}" != "0" ]; then
   confirm "You have uncommitted local changes. Push committed changes anyway?"
-  if [ "${?}" != "0" ]; then exit ${?}; fi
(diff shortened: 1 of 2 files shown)


git/hooks/pre-push
```diff
@@ -6,6 +6,15 @@ source "${REPO_DIR}/util/logging.sh"
 source "${REPO_DIR}/util/path.sh"
 WPTD_PATH=${WPTD_PATH:-$(absdir "${REPO_DIR}")}
 
+
+# Check for uncommitted changes.
+! git diff . 2>&1 | read > /dev/null 2>&1
+DIFF_STATUS="${?}"
+if [ "${DIFF_STATUS}" != "0" ]; then
+  confirm "You have uncommitted local changes. Push committed changes anyway?"
+  if [ "${?}" != "0" ]; then exit ${?}; fi
```
mdittmer: nit: Final `exit` in this script quotes the return code. I don't care whether it gets quoted (it's always a number), but we should make the file consistent.

lukebjerring: Quoted.

PR #310, opened 2017-11-28
Prototype for viewing a diff on wpt.fyi

lukebjerring:
Add-on for https://github.com/w3c/wptdashboard/issues/301

Running at https://diff-visualizer-dot-wptdashboard.appspot.com

Example:
https://diff-visualizer-dot-wptdashboard.appspot.com/?before=edge@1db039e5b6&after=edge


lukebjerring:
Considerations:
* Browser summary should probably be extracted, display something more useful than Diff@diff
* Diff is omitting entries for test-counts with unchanged results, which makes the numbers look strange.
* Existing UI uses redness shade inversely proportional to number of differences, which makes no sense
* Super slow to load

rwaldron:
@lukebjerring ping for rebase—thanks!

lukebjerring:
@rwaldron sync'd.

rwaldron:
@lukebjerring Github is still reporting that there are conflicts, which I can confirm by pulling this locally. Can you run this in your local branch: 

```
# assuming "upstream" is https://github.com/w3c/wptdashboard.git
git pull --rebase upstream master
```

Resolve the conflicts, commit, then...

```
git push HEAD -f
```

lukebjerring:
Not sure what you're seeing, but I see:

    This branch has no conflicts with the base branch
    Merging can be performed automatically.

rwaldron:
I just nuked everything local and started fresh and now I cannot reproduce this conflict that Github claims, so we're back on track.

components/runs.html
```diff
@@ -41,7 +41,7 @@
         async connectedCallback () {
           super.connectedCallback()
 
-          if (!this.testRuns) {
+          if (this.testRunResources) {
```
rwaldron: I'm trying to understand this change—I don't see any other changes that involve `this.testRunResources`. Any details would be appreciated 👍

components/runs.html
```diff
@@ -57,7 +57,7 @@
                 .reduce((sum, item) => {
                   return Array.isArray(item) ? sum.concat(item) : [...sum, item]
                 }, [])
-            this.testRuns = flattened
+            this.testRuns = this.testRuns ? [...flattened, ...this.testRuns] : flattened
```
rwaldron:
```js
this.testRuns = flattened.concat(this.testRuns || [])
```

wdyt?

components/wpt-results.html
```diff
@@ -364,6 +364,14 @@
               : '&'
           path += 'sha=' + this.sha
         }
+        var diffRuns = this.testRuns.filter(r => r['revision'] == 'diff')
```
rwaldron: Use MemberExpression.IdentifierName: `r.revision == 'diff'`

components/wpt-results.html
```diff
@@ -364,6 +364,14 @@
...
+        if (diffRuns.length > 0) {
+          path += path.indexOf('?') < 0
+                  ? '?'
+                  : '&'
+          path += 'before=' + this.testRuns[0]['browser_name'] + '@' + this.testRuns[0]['revision']
+          path += '&after=' + this.testRuns[1]['browser_name'] + '@' + this.testRuns[1]['revision']
```
rwaldron: - use destructuring to unpack `this.testRuns[0]` and `this.testRuns[1]`: 
    ```js
    let [before, after] = this.testRuns;
    ```
- use MemberExpression.IdentifierName (for all cases here and throughout): `before.browser_name`, `before.revision`, `after.browser_name` & `after.revision` 
- use template string: 
    ```js
    path += `before=${before.browser_name}@${before.revision}`
    ```

rwaldron:
Have you seen this?

lukebjerring:
Looks like the doubling-up of runs is already a bug - will fix separately.

rwaldron:
> doubling-up of runs is already a bug 

I'm not seeing this on master—where are you seeing it?

lukebjerring:
Sorry, edited comment to clarify what I meant.

components/runs.html
lukebjerring: TestRun is an array of the JSON objects, which can be passed in directly. TestRunResources are URLs to fetch the JSON objects. This PR uses both (the TestRun for the diff is passed as a JSON object).

This will probably change anyway, since the single-result view 'diff' isn't needed, so will probably keep 'diff' as a separate property on the Polymer element.

components/wpt-results.html
lukebjerring: TIL about unpack.

components/runs.html
rwaldron: Got it, thanks for the details!

components/test-file-results.html
```diff
@@ -144,6 +145,9 @@
       }
 
       resultsURL (testRun, testFile) {
+        if (testRun.revision == 'diff') {
+          return testRun.results_url + '&path=' + encodeURI(testFile)
```
rwaldron: Missed a template string update here.

components/wpt-results.html
```diff
@@ -364,13 +364,14 @@
...
+        var diffRuns = this.testRuns.filter(r => r.revision == 'diff')
         if (diffRuns.length > 0) {
           path += path.indexOf('?') < 0
                   ? '?'
                   : '&'
-          path += 'before=' + this.testRuns[0]['browser_name'] + '@' + this.testRuns[0]['revision']
-          path += '&after=' + this.testRuns[1]['browser_name'] + '@' + this.testRuns[1]['revision']
+          let [before, after] = this.testRuns
+          path += `before=${before.browser_name}@${before.revision}`
+          path += `&after=${after.browser_name}@${after.revision}`
```
rwaldron: 👍

PR #315, opened 2017-12-05
Move Google App Engine app to gae/

mdittmer:
This will help isolate the app from other infrastructure parts, and facilitate app + metrics infra dependencies as follows:

`github.com/w3c/wptdashboard/gae` depends on:
`github.com/w3c/wptdashboard` and `github.com/w3c/wptdashboard/metrics` for base and metrics data models

`github.com/w3c/wptdashboard/metrics` depends on:
`github.com/w3c/wptdashboard` for base data models

Notice that without `gae` sub-package this would introduce a circular dependency. The alternative to this change would be to create a `models` sub-package that both the base and `metrics` packages can depend on for base data models. However, that does not isolate web app code as nicely.

//cc @jeffcarp


mdittmer:
CI probably failing because this depends on #313

rwaldron:
#313 is landed, so you can rebase this now (ping me when ready)

rwaldron:
Needs a rebase—thanks!

mdittmer:
*ping*

lukebjerring:
`gcloud app deploy` is currently broken because of the subdirectories inside the service directory (which is currently the root). AppEngine doesn't treat github.com/w3c/wptdashboard as needing getting, so the metrics/ and other subfolders don't have the dependency resolved.

    ERROR: (gcloud.app.deploy) Error Response: [9] Deployment contains files that cannot be compiled: Compile failed:
    2017/12/14 08:17:14 go-app-builder: build timing: 5×compile (1.822s total), 0×link (0s total)
    2017/12/14 08:17:14 go-app-builder: failed running compile: exit status 2

    metrics/models.go:20: can't find import: "github.com/w3c/wptdashboard"
    metrics/compute/compute.go:21: can't find import: "github.com/w3c/wptdashboard"

I suspect this PR would resolve the problem, as the GAE and service (app.yaml) code would be in its own package?

mdittmer:
@rwaldron please review and merge this **before** any other golang changes. The PR keeps thrashing whenever other golang changes get merged to master.

rwaldron:
lol @ me... What I meant was: 

```
wptdashboard
.
├── metrics
│   └── metrics stuff
├── runner
│   └── The stuff presently in "run"
├── shared
│   └── models, etc.
└── wpt.fyi
    └── the google app engine stuff here
```

rwaldron:
Moving forward, please use `rebase` when feature branches fall behind master.

mdittmer:
I think using `.` (a convention for file extensions) is a bad idea. Moving `run` seems out of scope for this change. Moving golang code in `/` to `/shared` seems reasonable as a part of this change.

**Proposed compromise**: Add to this PR moving golang `/` to `/shared`. Leave `runner` move to a later PR. I'm open to non-`gae` suggestions for the main directory affected by the PR, but I'm not convinced `wpt.fyi` is a good idea.

Thoughts?

rwaldron:
> Add to this PR moving golang / to /shared. 

Can you clarify which files you mean, specifically?

> Leave runner move to a later PR. 

Yeah, that's fine. It's just a renaming, so easy to do in a follow up.

> I'm open to non-gae suggestions for the main directory affected by the PR, but I'm not convinced wpt.fyi is a good idea.


I picked `wpt.fyi` because it's simultaneously succinct and very explicit—there will be no question about the code in that directory. How about `wpt-fyi` or `wptd-app`?

mdittmer:
> > Add to this PR moving golang / to /shared. 

> Can you clarify which files you mean, specifically?

    mkdir shared
    git mv *.go shared/
    # refactor golang imports accordingly

What about `wptd-webapp` for the new folder name?

rwaldron:
> ```
> mkdir shared
> git mv *.go shared/
> # refactor golang imports accordingly
> ```

👍 

> What about wptd-webapp for the new folder name?

wfm

mdittmer:
> > What about wptd-webapp for the new folder name?

> wfm

**sigh** so golang doesn't like package names that contain dashes. How about `webapp`?

mdittmer:
*Aside about rebase on github*: I'm pretty sure rebasing when there are conflicts requires rewriting hashes, which nukes inline code review comments because the author has to `git push --force` the rewritten changes.

Personally, I also prefer the rebase strategy, but AFAICT it is incompatible with any kind of *ongoing code review* flow on github.com.

rwaldron:
> sigh so golang doesn't like package names that contain dashes. How about webapp?

Also works for me.

PR #329, opened 2018-01-05
Treat failed summary-fetch responses as empty JSON

lukebjerring:
This prevents a failure to parse (in the .json() call), which allows the other items to load in the UI correctly.

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -266,7 +266,7 @@ found in the LICENSE file.
 
       async _fetchResults (url) {
         const response = await window.fetch(url);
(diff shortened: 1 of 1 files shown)


rwaldron:
I've restarted the travis job that failed, my guess is that it was an erroneous failure

lukebjerring:
I've actually tried a re-run already, to no avail; the test passes locally @ master, so need to dig into why it's failing.

rwaldron:
Merged https://github.com/w3c/wptdashboard/commit/3dfa021e386829206faf12cbdcaf48d351647270

PR #331, opened 2018-01-05
Don't cap the git depth at 50.

lukebjerring:
A better solution is to change that particular unit-test to execute a fetch, but this is the quick-fix to repair the build.

PR #333, opened 2018-01-08
Run an unshallow fetch during test

lukebjerring:
Better fix than https://github.com/w3c/wptdashboard/pull/331 for the currently-broken build.


boazsender:
I like encapsulating this to the test that needs it more than the other approach in #331 :+1:

PR #334, opened 2018-01-09
Abstract storage interface and push results to GCS + BQ

mdittmer:
The interface is now cleaner and allows for multiple implementations. This is illustrated by providing and leveraging a BigQuery implementation (in addition to the existing Datastore-metadata + CloudStorage-data implementation).

### DIFF
--- a/metrics/models.go
+++ b/metrics/models.go
@@ -176,15 +176,24 @@ type TestRunStatus struct {
 	Status CompleteTestStatus
 }
 
-type MetricsRun struct {
-	StartTime *time.Time    `json:"start_time"`
-	EndTime   *time.Time    `json:"end_time"`
-	TestRuns  *TestRunSlice `json:"test_runs"`
+type PassRateMetadata struct {
+	StartTime time.Time      `json:"start_time"`
+	EndTime   time.Time      `json:"end_time"`
+	TestRuns  []base.TestRun `json:"test_runs"`
+	DataUrl   string         `json:"url"`
+}
+
+type FailuresMetadata struct {
+	StartTime   time.Time      `json:"start_time"`
+	EndTime     time.Time      `json:"end_time"`
+	TestRuns    []base.TestRun `json:"test_runs"`
+	DataUrl     string         `json:"url"`
+	BrowserName string         `json:"browser_name"`
 }
 
 // Output type for metrics: Include runs as metadata, and arbitrary content
 // as data.
 type MetricsRunData struct {
-	MetricsRun *MetricsRun `json:"metrics_run"`
-	Data       interface{} `json:"data"`
+	Metadata interface{} `json:"metadata"`
+	Data     interface{} `json:"data"`
 }
--- a/metrics/models.go
+++ b/metrics/models.go
@@ -176,6 +176,10 @@ type TestRunStatus struct {
 	Status CompleteTestStatus
 }
 
+// Metadata capturing:
+// - When metric run was performed;
+// - What test runs are part of the metric run;
+// - Where the metric run results reside (a URL).
 type PassRateMetadata struct {
 	StartTime time.Time      `json:"start_time"`
 	EndTime   time.Time      `json:"end_time"`
@@ -183,6 +187,11 @@ type PassRateMetadata struct {
 	DataUrl   string         `json:"url"`
 }
 
+// Metadata capturing:
+// - When failures report was gathered;
+// - What test runs are part of the failures report;
+// - Where the failures report resids (a URL);
+// - What browser is described in the report.
 type FailuresMetadata struct {
 	StartTime   time.Time      `json:"start_time"`
 	EndTime     time.Time      `json:"end_time"`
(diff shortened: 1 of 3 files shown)


metrics/models.go
```diff
@@ -176,15 +176,24 @@ type TestRunStatus struct {
 	Status CompleteTestStatus
 }
 
-type MetricsRun struct {
-	StartTime *time.Time    `json:"start_time"`
-	EndTime   *time.Time    `json:"end_time"`
-	TestRuns  *TestRunSlice `json:"test_runs"`
+type PassRateMetadata struct {
```
lukebjerring: Add a descriptive doc comment, for PassRateMetadata and FailuresMetadata both

metrics/run/collect_metrics.go
```diff
@@ -138,89 +143,199 @@ func main() {
...
 	outputBucket := gcsClient.Bucket(*outputGcsBucket)
-	outputCtx := storage.Context{
-		Context: ctx,
-		Client:  gcsClient,
-		Bucket: &storage.Bucket{
-			Name:   outputGcsBucket,
-			Handle: outputBucket,
+	datastoreClient, err := datastore.NewClient(ctx, *projectId)
+	if err != nil {
+		log.Fatal(err)
```
lukebjerring: Surprised we're not returning here - won't datastoreClient be nil or unusable in case of error?

metrics/run/collect_metrics.go
```diff
@@ -138,89 +143,199 @@ func main() {
...
+		for _, err := range errs {
+			log.Printf("Upload error: %v", err)
 		}
-		err := storage.UploadMetricsRunData(&outputCtx, &objName,
-			&totalsSummary)
-		if err != nil {
-			log.Println(err)
-			outputErrs <- err
+		if len(errs) > 0 {
+			log.Fatal(err)
```
lukebjerring: You've fallen out of the loop here, so err is back to the function-scoped var?

metrics/run/collect_metrics.go
mdittmer: If I understand correctly, `log.Fatal*()` functions terminate the process. Do you think it's clearer to also return?

metrics/run/collect_metrics.go
mdittmer: You're right. This should be `errs[len(errs)-1]` to get the most recent of the upload errors.

metrics/run/collect_metrics.go
lukebjerring: > Fatal is equivalent to Print() followed by a call to os.Exit(1).

TIL.

rwaldron:
@mdittmer did you see @lukebjerring's question here: https://github.com/w3c/wptdashboard/pull/334#pullrequestreview-87554758 ?

mdittmer:
Thanks @rwaldron; too focused on inline comments. @lukebjerring were you hoping for looser coupling at the `main()`-level? E.g., two implementations (one `results => GCS`, one `results => BQ`)? Or did you just want to make sure the description matched the content?

lukebjerring:
That's fine - updated PR description is better.

rwaldron:
@mdittmer & @lukebjerring same question as #335: How can this be tested/reviewed? Specifically: how can I observe the effect of these changes?

mdittmer:
https://metrics5-dot-wptdashboard.appspot.com/metrics/ is backed by data generated using this code. It's not really safe to write integration tests against this code without a staging server; the code is virtually all interface implementations that talk to remote services (and neither GCS nor BQ have local emulators :crying_cat_face: ).

mdittmer:
Here's how one would test with an empty staging project, `staging-wptd`:

1. Create a GCS bucket owned by `staging-wptd`, say `staging-wptd-metrics`
1. Ensure that BigQuery is enabled and functional on `stging-wptd`
1. Use flags to control where `collect_metrics.go` will send its output:
  1. `--project_id=staging-wptd`
  1. `--output_gcs_bucket=staging-wptd-metrics`
1. *Aside*: This configures output only; `wpt.fyi` endpoints and a `wptdashboard`-owned GCS bucket will still be the default location for inputs: latest test runs, results from latest test runs (resp.)
1. Compile and run `metrics/run/collect_metrics.go` with the above mentioned flags

The result should be that `gs://staging-wptd-metrics` and `staging-wptd`'s BigQuery datasets+tables are populated with metrics.

IMHO, because nearly all the code here is "the implementation of how to talk to underlying services", there is little value to writing elaborate stubs for the services and then testing those stubs.

rwaldron:
@mdittmer thank you for the clarification here

rwaldron:
@lukebjerring Are you +1?

lukebjerring:
Yes, +1 - I have a few structural (cosmetic) opinions, but I'm also of the opinion that with a pending chain of 3 PRs that are all out of sync, submitting in this state and doing a little file rearrangement afterwards is the best step forward. Ultimately, the code looks correct to me.

PR #336, opened 2018-01-10
Add some comments to the wpt.fyi paths

lukebjerring:
Improves documentation of the endpoints that exist in the one place that they're exhaustively listed.


rwaldron:
> I'm a little worried about this getting out of date. I don't have a great solutions. Cross-reference file names in both places?

If someone contributes code to the handlers, it's reasonable to expect them to update the explanations of those handlers.

PR #337, opened 2018-01-10
Give names to (and add) interop views

lukebjerring:
mdittmers [metrics demo](https://metrics5-dot-wptdashboard.appspot.com/metrics/) is one of two views from the [metrics design](https://docs.google.com/document/d/1MZRF3EdpMG0pLNI99jEj-pBITbqWKYMyJC7Ga8NtZ88/edit), which intends to add a view to show tests broken down by the number of browsers they pass in.

There are two views:
  - a list of tests that pass in some, but not all, browsers
  - the view demo'd above, i.e. tests broken down by interoperability

This PR adds placeholder endpoints for these views, and the main objective of the PR is to pick a name for each view.

### DIFF
--- a/webapp/anomaly_handler.go
+++ b/webapp/anomaly_handler.go
@@ -1,0 +1,20 @@
+// Copyright 2017 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package webapp
+
+import (
+	"net/http"
+)
+
+// anomalyHandler handles the view of test results showing which tests pass in
+// some, but not all, browsers.
+func anomalyHandler(w http.ResponseWriter, r *http.Request) {
+	// Empty struct placeholder.
+	data := struct{}{}
+	if err := templates.ExecuteTemplate(w, "anomalies.html", data); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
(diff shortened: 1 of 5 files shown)


foolip:
> [metrics design]

Intended to link to https://docs.google.com/document/d/1MZRF3EdpMG0pLNI99jEj-pBITbqWKYMyJC7Ga8NtZ88/edit?

> a list of tests that pass in some, but not all, browsers

Per chat, that's a union of the per-browser reports, merged into a single thing which can then be filtered. Maybe call it "Failing tests"? Could imagine later adding filtering by path to make it useful for spec people too.

> the view demo'd above, i.e. tests broken down by interoperability

Since it'll be the front page just calling it "Front page" would work I think, otherwise "Interop"?

foolip:
Actually "Failing tests" isn't going to be strictly true. There will be some testharness.js tests which actually didn't show up in some browsers because of how the tests are written, and those aren't failing. And "non-passing tests" doesn't quite have the same ring to it.

lukebjerring:
Yeah - I like 'anomalies' because it's not implying some other meaning that pass/fail have when running the tests in wpt. Interop failure != test failure

foolip:
Anomalies SGTM.

rwaldron:
@lukebjerring @foolip it appears that the main objective has been reached:

> This PR adds placeholder endpoints for these views, and the main objective of the PR is to pick a name for each view.

Merging!

foolip:
Hah, yesterday I didn't even notice this was a PR with suggestions built into it, I thought it was a discussion issue. Pardon my suggestions :)

PR #345, opened 2018-01-11
Various fixups to webapp/components/* files

rwaldron:
- There's no semantic value to prefixing methods with `_`. If privacy, or capability control is necessary, then a computed property (in the ES sense) using a Symbol is the appropriate pattern. 
- Reorganized class definitions such that methods are defined in this order: `compute*`, `async ...`, miscellaneous. 
- Removed unused code
- Eliminated unnecessary wrapper method
- Always use blocks

### DIFF
--- a/webapp/components/test-file-results.html
+++ b/webapp/components/test-file-results.html
@@ -54,8 +54,8 @@ found in the LICENSE file.
             <td class="sub-test-name">[[subtestName]]</td>
 
             <template is="dom-repeat" items="{{testRuns}}" as="testRun">
-              <td class$="result [[ _subtestResultForTestRun(testRun, subtestName) ]]">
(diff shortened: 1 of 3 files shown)


mdittmer:
IIRC `_methodame()` is an idiom inherited from Polymer. I'm not sure
whether Polymer treats such methods differently. @jeffcarp or @lukebjerring
do you know?

boazsender:
I just pulled ran, and tested locally. The app, and this view in particular, is behaving consistently with master. Merging in now.

@rwaldron In addition, changes look great, much more approachable with the blocks.

@mdittmer thanks for raising that point. I believe that was in a older version of polymer. In general, the `_` prefix was a broader JavaScript convention prior to the wider availability of access control in JS.

jeffcarp:
:+1: the underscore methods were via convention, Polymer doesn't treat them differently

PR #346, opened 2018-01-11
run/run.py: support providing an explicit wpt commit sha

rwaldron:
Supporting work for: 

> KR: wpt.fyi always shows results for the same wpt commit (P2)

This allows a remote test runner machine to specify which commit SHA of wpt should be run. Next, we need to decide on how to establish the "source of truth", ie. _what_ determines that SHA and then _where_ that information lives. 

To run the tests directly: 

```
docker exec -it -u $(id -u $USER):$(id -g $USER) wptd-dev-instance python run/run_test.py
```

### DIFF
--- a/run/run.py
+++ b/run/run.py
@@ -94,44 +94,34 @@ def main(platform_id, platform, args, config):
     print('==================================================')
     print('Setting up WPT checkout')
 
-    wpt_setup_commands = [
-        ['git', 'reset', '--hard', 'HEAD'],  # For wpt.patch
-        ['git', 'checkout', 'master'],
-        ['git', 'pull'],
-        ['./wpt', 'manifest', '--work'],
-    ]
-    for command in wpt_setup_commands:
-        return_code = subprocess.check_call(command, cwd=config['wpt_path'])
-        assert return_code == 0, (
-            'Got non-0 return code: '
-            '%d from command %s' % (return_code, command))
+    wpt_sha = setup_wpt(args, platform, config, logger)
 
-    patch_wpt(config, platform)
-
-    wpt_path = config['wpt_path']
-    sha_finder = shas.SHAFinder(logger)
-    WPT_SHA = (sha_finder.get_todays_sha(wpt_path)
-               or sha_finder.get_head_sha(wpt_path))
-    print('Current WPT SHA: %s' % WPT_SHA)
+    print('Current WPT SHA: %s' % wpt_sha)
 
     return_code = subprocess.check_call(
-        ['git', 'checkout', WPT_SHA], cwd=config['wpt_path'])
+        ['git', 'checkout', wpt_sha], cwd=config['wpt_path'])
     assert return_code == 0, (
         'Got non-0 return code: '
         '%d from command %s' % (return_code, command))
 
-    SHORT_SHA = WPT_SHA[0:10]
+    short_wpt_sha = wpt_sha[0:10]
 
-    LOCAL_REPORT_FILEPATH = "%s/wptd-%s-%s-report.log" % (
-        config['build_path'], SHORT_SHA, platform_id
+    abs_report_log_path = "%s/wptd-%s-%s-report.log" % (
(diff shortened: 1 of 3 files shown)


run/run.py
```diff
@@ -94,44 +94,34 @@ def main(platform_id, platform, args, config):
...
-
-    wpt_path = config['wpt_path']
-    sha_finder = shas.SHAFinder(logger)
-    WPT_SHA = (sha_finder.get_todays_sha(wpt_path)
-               or sha_finder.get_head_sha(wpt_path))
-    print('Current WPT SHA: %s' % WPT_SHA)
+    print('Current WPT SHA: %s' % wpt_sha)
 
     return_code = subprocess.check_call(
-        ['git', 'checkout', WPT_SHA], cwd=config['wpt_path'])
```
lukebjerring: What's the aversion to UPPERCASE (used to imply 'constant' values)?

run/run.py
```diff
@@ -254,6 +244,29 @@ def main(platform_id, platform, args, config):
     print('Response text:', response.text)
 
 
+
+def setup_wpt(mainargs, platform, config, logger):
```
lukebjerring: Can you add some PEP484 type annotations?

mdittmer:
Drive-by review: Is `run.py` now used by running infra? I thought it was replaced by largely-duplicate `jenkins.py`.

run/run.py
rwaldron: I can, but maybe that should occur in a follow up PR? There are no such annotations elsewhere in this file, but it would be nice to add them for _all_ functions in this file.

foolip:
Was the "Various fixups to webapp/components/* files" commit really intended to be part of this PR? Seems quite unrelated.

run/run.py
```diff
@@ -254,6 +244,29 @@ def main(platform_id, platform, args, config):
...
+        assert return_code == 0, (
+            'Got non-0 return code: '
+            '%d from command %s' % (return_code, command))
+
+    patch_wpt(config, platform)
+
+    if mainargs.wpt_sha:
+        return mainargs.wpt_sha
+    else:
+        sha_finder = shas.SHAFinder(logger)
```
foolip: Do you think that this code path should later be retired by always requiring the sha to be provided? We'll end up here if there's a typo on the command line, so maybe emitting a warning that we picked a commit would make sense. Or a TODO to retire this stuff.

run/run.py
```diff
@@ -254,6 +244,29 @@ def main(platform_id, platform, args, config):
...
+        return (sha_finder.get_todays_sha(config['wpt_path'])
+                or sha_finder.get_head_sha(config['wpt_path']))
```
foolip: FWIW, this bit won't make sense if the `git checkout master; git pull` is changed. It'd work if `get_head_sha` did `git rev-parse origin/HEAD` instead of `git rev-parse HEAD`.

rwaldron:
@foolip 

> Was the "Various fixups to webapp/components/* files" commit really intended to be part of this PR? Seems quite unrelated.

Definitely not... I'm not sure what the hell happened there 0_o. I will sort that out when I address review stuff here.

rwaldron:
@foolip just needed a `git pull --rebase origin master` :)

run/run.py
```diff
@@ -254,6 +244,29 @@ def main(platform_id, platform, args, config):
     print('Response text:', response.text)
 
 
+def setup_wpt(mainargs, platform, config, logger):
+    wpt_setup_commands = [
+        ['git', 'reset', '--hard', 'HEAD'],  # For wpt.patch
+        ['git', 'checkout', 'master'],
```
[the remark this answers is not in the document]
rwaldron: I'll check that out—thanks for the tip :)

run/run.py
rwaldron: I'll take the "TODO" path for now, so I/we can think more about this

foolip:
@rwaldron, did you merge by accident? There were some TODOs to be sprinkled still.

Issue #350, opened 2018-01-12
Directories have links to themselves within that don't work (can't drill down)


foolip:
https://wpt.fyi/html now looks like this:
<img width="823" alt="screen shot 2018-01-12 at 9 01 21 pm" src="https://user-images.githubusercontent.com/498917/34893032-c5812284-f7db-11e7-887d-11ceda738182.png">

https://wpt.fyi/html/dom (which I knew existed) instead looks like this:
<img width="820" alt="screen shot 2018-01-12 at 9 02 24 pm" src="https://user-images.githubusercontent.com/498917/34893072-e8f36fb0-f7db-11e7-94a9-efce7e22b76b.png">

I don't know what the numbers mean here, but there is no html/ directory there, but are other directories:
https://github.com/w3c/web-platform-tests/tree/master/html/dom

This makes wpt.fyi currently unusable.

mattl:
Do we have a sense of when this *was* working?

foolip:
It was working when I filed https://github.com/w3c/wptdashboard/issues/344

rwaldron:
@foolip I've reverted f39dbaa5c4585fd02da86ceea1a5039156acb0a5 for now, since that was the most likely candidate for introducing site facing issues.

rwaldron:
@foolip I determined the cause of the issue, and opened a PR with webcomponents fixups _and_ a commit that fixes the regression.

boazsender:
This was resolved by #351.

foolip:
Thanks for the speedy fix everyone!

PR #351, opened 2018-01-12
Various fixups to webapp/components/* files, Revised!

rwaldron:
This contains the same commit that was previously landed and reverted, as well as a revision commit that "reverts" the dir/file drill down display code.


boazsender:
I just pulled and tested this and it resolves the issue that @foolip caught in #350. I apologize for missing that the first time.

Merging now.

PR #353, opened 2018-01-12
fix docs for deployment

boazsender:


docs/app-engine.md
```diff
@@ -17,6 +17,7 @@ Make sure you have the latest code and deploy the app.
 
 ```sh
 git pull
+cd webapp/
 gcloud app deploy
```
lukebjerring: Can we change the standard deploy to set the AppEngine version flag to the short-sha (or long?).

`gcloud app deploy -v=$(git rev-parse --short HEAD)`

This would make it a lot easier to tell what git state is running for a given 'version' in AppEngine, and makes rolling back easier + safer.

lukebjerring: Docs for context on the --version flag: https://cloud.google.com/sdk/gcloud/reference/app/deploy

mdittmer: That approach can get dicey if you don't check for a clean working tree in the repo (and any unversioned deps, which I don't think we have any). [Something like this](https://github.com/GoogleChromeLabs/confluence/blob/master/scripts/deploy.sh) could take care of it.

rwaldron: @boazsender ^^^

lukebjerring:
Relevant to https://github.com/w3c/wptdashboard/pull/353

PR #357, opened 2018-01-15
Initial test for metrics/compute

mdittmer:


metrics/compute/compute_test.go
```diff
@@ -0,0 +1,80 @@
...
+				&testName,
+				[]metrics.SubTest{},
+			},
+		},
+	}
+	gathered := GatherResultsById(results)
+	assert.Equal(t, len(gathered), 1) // Merged to single TestId: {"A test",""}.
+	for testId, runStatusMap := range gathered {
+		assert.Equal(t, testId, metrics.TestId{"A test", ""})
+		for testRun, testStatus := range runStatusMap {
```
lukebjerring: Wouldn't it be easier to assert.Equal(t, 2, len(runStatusMap)) and then look up (by key) runA and runB?

metrics/compute/compute_test.go
```diff
@@ -0,0 +1,80 @@
...
+			&metrics.TestResults{
+				"A test",
+				"ERROR",
+				&testName,
+				[]metrics.SubTest{},
+			},
+		},
+	}
+	gathered := GatherResultsById(results)
+	assert.Equal(t, len(gathered), 1) // Merged to single TestId: {"A test",""}.
```
lukebjerring: Swap params @ 1 and 2 around - the API is (t, expected, actual)

metrics/compute/compute_test.go
mdittmer: Sure.

rwaldron:
Landed! a907032c8114bf4821df4c7fe60d202bacf071b1

Tested with: `docker exec -it -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make go_test`

PR #359, opened 2018-01-15
Metrics=>GCS: Write data, then update gzip metadata

mdittmer:
Without reversing the data-write+metadata-write ordering, the metadata-write fails (and hence metrics data collection fails).

### DIFF
--- a/metrics/storage/storage.go
+++ b/metrics/storage/storage.go
@@ -89,23 +89,27 @@ func (ctx GCSDatastoreContext) Output(id OutputId, metadata interface{},
 	log.Printf("Writing %s to Google Cloud Storage\n", name)
 	gcsData := GCSData{metadata, data}
 	obj := ctx.Bucket.Handle.Object(id.DataLocation.GCSObjectPath)
+	objWriter := obj.NewWriter(ctx.Context)
+	if err := func() error {
+		gzWriter := gzip.NewWriter(objWriter)
+		encoder := json.NewEncoder(gzWriter)
+		if err := encoder.Encode(gcsData); err != nil {
+			return err
+		}
+		gzWriter.Close()
+		return objWriter.Close()
+	}(); err != nil {
+		log.Printf("Error writing %s to Google Cloud Storage: %v\n",
+			name, err)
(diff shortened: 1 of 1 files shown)


lukebjerring:
Fixes https://github.com/w3c/wptdashboard/issues/360

metrics/storage/storage.go
```diff
@@ -89,6 +89,21 @@ func (ctx GCSDatastoreContext) Output(id OutputId, metadata interface{},
 	log.Printf("Writing %s to Google Cloud Storage\n", name)
 	gcsData := GCSData{metadata, data}
 	obj := ctx.Bucket.Handle.Object(id.DataLocation.GCSObjectPath)
+	objWriter := obj.NewWriter(ctx.Context)
+	if err := func() error {
+		gzWriter := gzip.NewWriter(objWriter)
+		encoder := json.NewEncoder(gzWriter)
+		if err := encoder.Encode(gcsData); err != nil {
+			return err
```
lukebjerring: If this errors, the writers won't close. The `defer` statements from the previous file-state should be kept.

If your intent is to have the write occur immediately, use flush:
https://golang.org/pkg/compress/gzip/#Writer.Flush

mdittmer: Hmm. We are also trying to return potential error from closing `objWriter`. PTAL at my change: I defer closing `gzWriter` and manually close `objWriter` (discarding potential error) if we encounter an encoding error.

metrics/storage/storage.go
```diff
@@ -92,11 +92,12 @@ func (ctx GCSDatastoreContext) Output(id OutputId, metadata interface{},
 	objWriter := obj.NewWriter(ctx.Context)
 	if err := func() error {
 		gzWriter := gzip.NewWriter(objWriter)
+		defer gzWriter.Close()
 		encoder := json.NewEncoder(gzWriter)
 		if err := encoder.Encode(gcsData); err != nil {
+			objWriter.Close()
```
lukebjerring: Use CloseWithError if the error's non-nil.

PR #368, opened 2018-01-17
components/wpt-results.html: use regexp to validate test file

rwaldron:


rwaldron:
I intentionally cancelled the builds on travis because there is nothing in this changeset that could possibly affect a different build outcome.

PR #375, opened 2018-01-18
Restore shas.SHAFinder mock

rwaldron:


rwaldron:
Looks like this fixed the py_test failure

lukebjerring:
https://github.com/w3c/wptdashboard/pull/377 uses the standard `patch` method to do the same thing.

PR #377, opened 2018-01-19
use unittest.mock.patch to undo ctor mocking

lukebjerring:
Patch is a built-in method for achieving the same thing.

### DIFF
--- a/run/run_test.py
+++ b/run/run_test.py
@@ -80,17 +80,14 @@ class TestRun(unittest.TestCase):
         run.patch_wpt = stub_patch_wpt
         subprocess.check_call = stub_check_call
 
-        originalSHAFinder = shas.SHAFinder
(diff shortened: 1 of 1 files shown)


rwaldron:
This was really unclear to me when I was reading the docs for mock 2.0.0. It's strange that only the _patch_ method has restore-ability.

lukebjerring:
@rwaldron I'm not sure why Travis didn't run for this PR. Will try and manually trigger a run (even if only to make it more obvious whether this caused a regression). Tests definitely passed locally.

rwaldron:
> Tests definitely passed locally.

I had the same experience—I ran the tests locally before merging. This is really strange.

lukebjerring:
Seeing whether it's an unrelated non-patch stub that prevents the mock getting called?
 https://github.com/w3c/wptdashboard/pull/386

PR #380, opened 2018-01-22
Host environment shell script for installing default application cred…

mdittmer:
…entials inside docker container


rwaldron:
Worked perfectly!

Issue #378, opened 2018-01-22
populate_dev_data.go not functioning correctly


lukebjerring:

lukebjerring:
Note that https://github.com/w3c/wptdashboard/pull/380 is a browser-based method for acquiring application credentials.

lukebjerring:
@rwaldron - Looks like this auto-closed via the "Fixes #378" comment in #381. Can you confirm that you're now able to run the script locally (including the instructions being reasonable)?

rwaldron:
Sorry I missed this—yes, this is functioning correctly now.

PR #381, opened 2018-01-22
Add script for using populate_dev_data.go in docker

lukebjerring:
Leverages the populate_dev_data.go in docker, to avoid issues (fixes https://github.com/w3c/wptdashboard/issues/378).

For compatibility with docker-external execution, this script leverages the `$GOOGLE_APPLICATION_CREDENTIALS` variable, copying the file into docker.


lukebjerring:
Note that https://github.com/w3c/wptdashboard/pull/380 is another docker-based auth flow, which requires the user to open a link in a browser (and is needed every time the server is reset).

Personally, I prefer this flow, which is well documented [in the AppEngine docs](https://developers.google.com/accounts/docs/application-default-credentials)

Issue #384, opened 2018-01-22
Fix use of mock.patch('shas.SHAFinder') 


rwaldron:

lukebjerring:
Fixed by https://github.com/w3c/wptdashboard/pull/386 and https://github.com/w3c/wptdashboard/pull/385

PR #366, opened 2018-01-16
wip: preliminary travis testing

rwaldron:
This is now green.

PR #385, opened 2018-01-22
Use mock.patch('shas.SHAFinder.get_todays_sha'). Fixes gh-384

rwaldron:


rwaldron:
This is really strange: the failure doesn't occur locally

rwaldron:
What's even more unsettling is that the previous code runs and passes with no issue on this branch: https://github.com/w3c/wptdashboard/pull/366 https://travis-ci.org/w3c/wptdashboard/builds/331966110

Issue #374, opened 2018-01-18
wpt.fyi/* routes are future hostile


rwaldron:
If https://github.com/w3c/web-platform-tests were to ever grow a new top level directory for some hypothetical spec named "api", "interop", "results", or "test-run", then routing in webapp/main.go becomes ambiguous.

boazsender:
+1, test results routes should be name spaced.

For the `/interop/` example in #372, it's worth mentioning that the [compat spec](https://compat.spec.whatwg.org) could have just as easily been called "interop". We have tests for that spec, and the `/compat` route is in the global name space (see https://wpt.fyi/compat).

rwaldron:
Honestly, the only truly safe solution is the one that breaks all the existing inbound links, ie. wpt.fyi/test/path.html => wpt.fyi/some-safe-path/test/path.html, but that might be the price to pay for rushing this out.

lukebjerring:
We can redirect the current wildcard matches to /some-safe-path/, mitigating breakages to current links to clashes only.

rwaldron:
@lukebjerring yep, that's what was I thinking (and what brought me back here just now to follow up on). I don't want to rush into this yet.

lukebjerring:
This issue is related to the Q1 OKR of switching the default view to the interop view:
https://docs.google.com/document/d/1ojlLDiRruI4e95Z7JQ10JQPRMiQHd_40rBkYCL6IJqw/edit#heading=h.v93ez0y7fvmy

lukebjerring:
This issue was moved to web-platform-tests/wpt.fyi#13

PR #387, opened 2018-01-22
Add path-part component

lukebjerring:
Step towards a consolidated URI path treatment on wpt.fyi for https://github.com/w3c/wptdashboard/issues/374.

Running at https://lukebjerring-path-part-dot-wptdashboard.appspot.com

path-part takes an optional prefix, where `/interop` will we used for interop, and eventually we will pick a prefix for the current default view (to be future-friendly - see the issue for more context).

### DIFF
--- a/webapp/components/path-part.html
+++ b/webapp/components/path-part.html
@@ -1,0 +1,78 @@
+<!--
+Copyright 2017 The WPT Dashboard Project. All rights reserved.
+Use of this source code is governed by a BSD-style license that can be
+found in the LICENSE file.
+-->
+
+<link rel="import" href="../bower_components/polymer/polymer-element.html">
+
+<!--
+`<path-part>` is a stateless component for displaying part of a test path.
+-->
+<dom-module id="path-part">
+  <template>
+    <style>
+      a:hover {
+        cursor: pointer;
+        color: #226ff3;
+      }
+      .dir-path {
+        font-weight: bold;
+      }
+    </style>
+
+    <a class$="{{ styleClass }}" href="{{ href }}" on-click="navigate">
+      {{ relativePath }}
+    </a>
+  </template>
+
+  <script>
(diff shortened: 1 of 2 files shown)


rwaldron:
I love this

webapp/components/path-part.html
```diff
@@ -0,0 +1,72 @@
...
+
+<link rel="import" href="../bower_components/polymer/polymer-element.html">
+
+<!--
+`<path-part>` is a stateless component for displaying part of a test path.
+-->
+<dom-module id="path-part">
+  <template>
+    <style>
+      .dir-path { font-weight: bold; }
```
rwaldron: Where is `.file-path`?

webapp/components/path-part.html
```diff
@@ -0,0 +1,72 @@
...
+    </style>
```
rwaldron: The styles here: https://github.com/lukebjerring/wptdashboard/blob/baf129bf3fbf6cfdc72820d3e4acb7aba8ff181f/webapp/components/wpt-results.html#L36-L44 have been lost.

webapp/components/path-part.html
lukebjerring: It's 'not bold' so doesn't have any extra styles.

webapp/components/path-part.html
lukebjerring: Restored - sorry that was sloppy; I was trying to minimize the repeated (shadow DOM) styles.

lukebjerring:
Rebased, styles restored, and redeployed (@ https://lukebjerring-path-part-dot-wptdashboard.appspot.com) - PTAL.

rwaldron:
Landed https://github.com/w3c/wptdashboard/commit/ea0ace1e139483f53b6f9e90512d1d9bf43a9f8f

PR #403, opened 2018-01-25
Remove /etc/{passwd,group} mounts which break docker 17.12.0-ce-mac46

lukebjerring:


lukebjerring:
Notes on this change:

Docker seems to no longer set extra file attributes; Before:

    $ xattr -l generated/strawman.pb.go
    com.docker.owner: 0:187761:5000

After deleting the file and re-running `docker exec -t -u $(id -u $USER):$(id -g $USER) wptd-dev-instance make proto`, the output of `xattr` is empty (and, the trailing `@` in `ls -al generated/` is gone).

The owner still appears to be me;

    -rw-r--r--   1 lukebjerring  eng   3.6K Jan 25 13:39 browser.pb.go

@mdittmer :- 

> if you do something that writes files, who owns them on the host machine?

Does this answer that concern? Is there any other need for it?

lukebjerring:
Also verfied the same thing on linux.

PR #407, opened 2018-01-26
Move styles for <path-part> into path-part.html

rwaldron:


lukebjerring:
The downside of this is that the style tag is repeated for every element. We only have ~20 items in a packed list though, so probably not worth optimizing for.

rwaldron:
One of the motivations for using web components is that they are self-contained entities. If we want to have a global style, then we should do that, but we should also eliminate the component specific styles.

PR #409, opened 2018-01-29
Fix interop path-part navigate + stale class

lukebjerring:
Same bug as https://github.com/w3c/wptdashboard/pull/405 - I didn't have a chance to tweak https://github.com/w3c/wptdashboard/pull/404 to include the same fix.


rwaldron:
Landed 35563fcc31fd09a360f8e9197e0a060a1e8e5bd4

Issue #416, opened 2018-01-31
Init (and update) git submodule automatically?


Hexcles:
#412 added git submodule to this repo. Albeit a common step, contributors would now need to run `git submodule update --init` themselves. Shall we perhaps automate the step somewhere? Or at least document it in README.md.

cc @lukebjerring

PR #417, opened 2018-01-31
Init submodule as part of Makefile rule

lukebjerring:
Fixes https://github.com/w3c/wptdashboard/issues/416

cc @Hexcles

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,20 +58,24 @@ go_lint: go_deps
 go_test: go_deps
 	cd $(WPTD_GO_PATH); go test -v ./...
 
-bq_proto: $(PROTOS)
+bq_proto: submodules $(PROTOS)
 	mkdir -p $(PB_BQ_OUT_DIR)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -58,21 +58,25 @@ go_lint: go_deps
...
 
-py_proto: $(PROTOS)
+py_proto: submodules $(PROTOS)
 	mkdir -p $(PB_PY_OUT_DIR)
 	protoc -I$(PB_LIB_DIR) -I$(PB_BQ_LIB_DIR) -I$(PB_LOCAL_LIB_DIR) \
 		--python_out=$(PB_PY_OUT_DIR) $(PROTOS)
 
+# Ensure submodule have been pulled.
+submodules:
+	cd $(WPTD_PATH); git submodule update --init
```
Hexcles: Just a nit: we probably don't need `cd $(WPTD_PATH)` as we are at this directory already.

PR #383, opened 2018-01-22
Derive and include a "commit_date" for new TestRun entities. Ref gh-154

rwaldron:
See #154 for context.

PR #426, opened 2018-02-03
Execute WPT in 'chunks' & attempt failure recovery

jugglinmike:


rwaldron:
Can I get a commitment to unit tests at some point in the future?

run/run.py
```diff
@@ -127,65 +133,96 @@ def main(platform_id, platform, args, config):
...
+            else:
+                sauce_browser_name = platform['browser_name']
+
+            command = [
+                './wpt', 'run', 'sauce:%s:%s' % (
+                    sauce_browser_name, platform['browser_version']),
+                '--sauce-platform=%s' % platform['os_name'],
+                '--sauce-key=%s' % config['sauce_key'],
+                '--sauce-user=%s' % config['sauce_user'],
+                '--sauce-connect-binary=%s' % config['sauce_connect_path'],
```
rwaldron: Since we're shaking things up... what do you think about changing `config['sauce_connect_path']` and its corresponding key in runner.ini to `sauce_connect_binary`? We can do that in another PR if you want.

rwaldron:
Surprisingly, this will apply cleanly to https://github.com/w3c/wptdashboard/pull/383

run/run.py
jugglinmike: That's a good idea. I've also noticed some inconsistency with the treatment of the Firefox and Chrome binaries. As you've pointed out, though, there are no tests for this patch, so I'm inclined to defer those improvements

jugglinmike:
> Can I get a commitment to unit tests at some point in the future?

Given the potential for churn in the implementation of `run.py` over the next
month or so, my preference is to write integration tests. Integration testing
a Python script like this is new territory for me, so we're tracking that work
with a dedicated issue: #428.

In the short term, we recognized that this modification has the potential to
greatly increase the completeness of test results. So even though, like you,
I'm reluctant to work without the benefit of tests, this order of operations is
a strategic decision, intentionally made.

rwaldron:
> Given the potential for churn in the implementation of run.py over the next
month or so, my preference is to write integration tests.

Works for me—thanks!

Issue #109, opened 2017-08-31
Test Edge Insider Preview


jeffcarp:
This is available from [Browserstack](https://www.browserstack.com/test-on-microsoft-edge-browser#live-cloud) or it could be set up on a VM if possible.

foolip:
I can only see Edge 15 on BrowserStack, shouldn't the pre-release versions now be 16?

jgraham:
https://www.browserstack.com/list-of-browsers-and-platforms?product=live doesn't list any dev version for Edge.

Given that Edge preview releases are tied to Windows versions, I don't know any realistic alternative to running our own hardware here. With some effort you could probably script setting up a VM from https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/ and use that to keep up to date with the latest available builds.

foolip:
FWIW, I did once try to install Windows Insider (not the Edge VMs) in VirtualBox and get it to upgrade to the most recent version, but never got it working. With a physical laptop I've succeeded, though.

gsnedders:
I feel like @thejohnjansen or @patrickkettner not say MS could potentially get us some reasonable way of having constantly up-to-date? From memory, trying to set up Insider builds for CI is virtually impossible because you need to leave the VM on for long enough to actually update (and presumably then save that state to create a new VM of the latest build), and it's very non-deterministic how long it takes before Windows Update even notices there's a new build (even if you literally tell it to repeatedly search for updates!).

foolip:
I don't know how much work this would require on @bobholt's side, but I think the setup would be roughly:
- Once a day, a URL like http://wpt.fyi/TEST_THESE_COMMITS is updated, so that everyone is testing using the same commit. (See https://github.com/GoogleChrome/wptdashboard/issues/40.)
- Microsoft runs the whole test suite, using the scripts checked into the web-platform-tests repo, or at least somewhere public so that it can be reproduced.
- Results are submitted to http://wpt.fyi/HERE_ARE_MY_RESULTS (with some auth thing) and show up on the dashboard.

foolip:
We've done some planning for Q4 now, and unfortunately this issue does not seem tractable right now. It's possible that BrowserStack or Sauce will get more recent versions of Edge, and there's been some discussion about Microsoft running tests and submitting builds. Neither of these seem imminent, so we'll revisit this for Q1.

@thejohnjansen @patrickkettner please shout loudly if this is not what you expected and we will do a 180!

[...]

PR #434, opened 2018-02-07
Remove "caveat" sections from top of page, create new footer

rwaldron:


webapp/static/common.css
```diff
@@ -1,51 +1,58 @@
...
+  font-size: 1.5em;
+  margin-bottom: 0.5em;
 }
 article p, ul {
-    margin-bottom: 1em;
+  margin-bottom: 1em;
 }
 article ul {
-    margin-left: 2em;
-}
```
rwaldron: All of these changes are whitespace/indentation.

webapp/templates/_header.html
```diff
@@ -7,4 +7,7 @@ <h1><a href="/">web-platform-tests dashboard</a></h1>
     <a href="/about">About</a>
     <a href="https://github.com/w3c/wptdashboard">GitHub Source</a>
   </nav>
+  <section class="caveat">
+    This site is a work-in-progress.
```
rwaldron: @jugglinmike please advise here

webapp/static/common.css
jugglinmike: In the future, a dedicated commit would be nice (as a GitHub.com comment your explanation won't show up in the git history). I don't think it's worth the effort to change it here, though

webapp/templates/_footer.html
```diff
@@ -0,0 +1,6 @@
+<footer>
+  <section class="caveat">
+    This information is intended for web platform implementers and does not contain
+    useful metrics for evaluation or comparison of web platform features.
```
jugglinmike: How about:

"The Web Platform Tests Dashboard is a work-in-progress. The reported results do not necessarily reflect the true capabilities of each web browser."

webapp/static/common.css
```diff
@@ -1,51 +1,58 @@
...
\ No newline at end of file
+  margin-left: 2em;
+}
+.caveat {
+  margin-top: 20px;
+  padding: 0.5em;
```
jugglinmike: Could we stick with `em` units for consistency's sake?

webapp/templates/_footer.html
jugglinmike: Oh, err, I think maybe the text I'm recommending belongs in the header.

webapp/static/common.css
rwaldron: Noted!

webapp/static/common.css
rwaldron: lol, sorry—that was lazy ;)

webapp/templates/_footer.html
rwaldron: I've updated the top banner :)

jugglinmike:
Now that I've finally run this locally and seen the visual design, I have some additional feedback on the copy itself. It's a lot of text, and its distinctive visual treatment is a bit jarring when presented across three elements.

Simply combining the messages is probably not a good idea, since that would result in a lengthy paragraph above-the-fold. That paragraph also wouldn't read so well because each sentence was written independently.

What do you think about re-writing the copy to be more terse and including that shorter text in the header and footer of every page?

More concretely, I'm thinking we can go from this:

> > The Web Platform Tests Dashboard is a work-in-progress. The reported results do not necessarily reflect the true capabilities of each web browser.
>
> [...]
>
> > Data below are intended for web platform implementers and do not contain useful metrics for evaluation or comparison of web platform features. Each browser listed here is currently tested at its latest stable release. See #109 and #112 for discussion on testing pre-release versions. 
>
> [...]
>
> > This information is intended for web platform implementers and does not contain useful metrics for evaluation or comparison of web platform features. 

To something like this:

> The Web Platform Tests Dashboard is a work-in-progress. The reported results do not necessarily reflect the true capabilities of each web browser, so they should not be used evaluate or compare feature support. Each browser listed here is currently tested at its latest stable release.

(Apologies for the delay on this feedback--I spent the morning struggling with the proprietary Google technologies that the project depends on.)

rwaldron:
@jugglinmike updates made in new commit

rwaldron:
Merged

jugglinmike:
Sorry, I'm still new to the project. How do these changes get deployed?

rwaldron:
@jugglinmike I can understand how that might be frustrating, considering there is almost no documentation available in this repository.

PR #450, opened 2018-02-12
Remove superflous banner

jugglinmike:
Recent changes in the surrounding interface have obviated the need for the caveat in this banner element.


rwaldron:
I think I meant to remove this in https://github.com/w3c/wptdashboard/pull/434

rwaldron:
👍

Issue #444, opened 2018-02-09
Update: remove xvfb from script (was: xvfb should be optional)


mattl:

jugglinmike:
The problem here is that we don't want to use `xvfb` in all cases. An additional flag could get us there, but there are other approaches we should consider.

For example, what about removing `xvfb-run` from `run.py` altogether? This would make `run.py` more portable overall. As far as I can tell, portability is the primary motivation for scripting in Python in the first place. That would let us use the same executable to run Edge locally (as opposed to remotely via Sauce Labs). We may have a similar need for virtual displays in Windows environments, but it won't be solved by `xvfb`.

mattl:
I'm okay with that. We could leave providing a viable running solution to the machine itself, and perhaps use a headless mode if one exists as a last resort. I'll close out my PR for now.

jugglinmike:
We addressed this issue via the introduction of a Buildbot-driven infrastructure (see gh-522).

PR #462, opened 2018-02-14
Adds an option to ignore xvfb

mattl:
This solves #444

### DIFF
--- a/run/run.py
+++ b/run/run.py
@@ -153,11 +153,14 @@ def main(platform_id, platform, args, config):
             if args.path:
                 command.insert(3, args.path)
         else:
-            command = [
+            command = [                    
                 'xvfb-run', '--auto-servernum',
                 './wpt', 'run',
                 platform['browser_name'],
             ]
+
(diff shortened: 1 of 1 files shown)


run/run.py
```diff
@@ -153,12 +153,15 @@ def main(platform_id, platform, args, config):
...
                 command.insert(3, args.path)
         else:
-            command = [
+            command = [                    
                 'xvfb-run', '--auto-servernum',
                 './wpt', 'run',
                 platform['browser_name'],
             ]
 
+            if args.no-xvfb:
```
jugglinmike: `no-xvfb` is not a valid Python property name. The expression will be interpreted as `args.no - xvfb`. That's inappropriate because `args` does not have a property named `no`, there is no binding named `xvfb`, and the result of a subtraction operation is not relevant for this code path.

We'll want to reference the property named `no_xvfb`, but before we can do that, we have to define the command-line option first.

mattl:
Closing as we're discussing removing xvfb-run from run.py. #444

Issue #292, opened 2017-11-22
Announce commit or commits to test for internal and external runs


foolip:
Background:
* The original [Design Doc: WPT Dashboard](https://docs.google.com/document/d/1eGsFVBaeg8q1LYowDj62EA-yXJgGuKX9A5FIkmejMxA/edit?usp=sharing) had "one build per platform every morning that runs all WPT tests against one WPT revision" and then, as a simplification, "Instead of bots pulling the WPT revision to test against, they just test against the first WPT revision that day".
* https://github.com/w3c/wptdashboard/issues/40 is about that simplification although the original central "pick a revision" part no longer exists.
* [Consistent results for all browsers are updated every 24 hours](https://docs.google.com/document/d/1GZquS8Jra47E5hIMH2ObOb_7oQFgk8f9MptUAjF0LFI/edit#heading=h.qxotds9ar227) will require results from the same commit. "Picking a WPT commit" is part of the description, although it's not strictly required to achieve the objective.
* https://github.com/w3c/wptdashboard/issues/164 is about testing much more frequently than once a day.
* https://github.com/w3c/wptdashboard/issues/290 is about supporting external runs.

Those are the "use cases" to solve. Proposed solution and title of this issue:

Have an endpoint like https://wpt.fyi/api/shas that returns at the very least *a* commit to test, but more usefully a set of commits which is either pre-prioritized or has enough information that consumers could decide which to prioritize running first.

The minimum useful behavior for runners would be to prioritize finishing a well-defined "today's sha".

foolip:
@mattl, would you consider this resolve by https://sha.wpt.fyi/? I think a bit more would be required to support external runs, but it's fine to keep that to https://github.com/w3c/wptdashboard/issues/290.

jugglinmike:
As of gh-522, we collect results on a regular schedule (once every 6 hours for Chrome and Firefox, once every 2 days for Edge and Safari). We identify the latest commit to WPT at this moment, and all subsequent testing operations reference that commit via its git SHA sum.

PR #463, opened 2018-02-14
sha.wpt.fyi

mattl:
This is a deliberately-simple HTML page and shell script used to 
publish a current working commit SHA at https://sha.wpt.fyi -- by
publishing the current working commit, browsers in and outside of the
wpt.fyi infrastructure can test against the same commit. Please send
me feature suggestions on how to improve things.


lukebjerring:
Note that following the link doesn't actually work; Can be 'fixed' with https://github.com/lukebjerring/wptdashboard/pull/3 , but I suspect you're intending to change the datastore to have full-length SHAs?

run/htdocs/index.html
```diff
@@ -0,0 +1,19 @@
...
...WVpaisvlIjFRsp2zgVDc9ufNz9m4aGOIowXeuO92pBq/xyD4M8BXVFTkswJLlizxIYguev4VR/52GYb7gG/Hjh0Rv+ctLS2+mJiYcO67nFLn9wx2XtFBuL6/UFhyWvtzhYO6ujrOnTsHQuFlPdqDalcNv6jIYSfwH3fs2KF2P40Udu3ahdvtBsE9+GOJIW7gZdDnd7qjyvcZNYBv27ZtEV1Z/vqv/1pZIexBz4ssFgK+oUOH+np6eiJ634uLi5X7/p8tvgfvEWH8LeD7wQ9+ENEHbvLkycoD968s/v5WwIGQUfLt3bs3Yvf82bNnPqfT6UN4UdmW3oH3iDiWAr4RI0b4+vv7I/LAXbhwQTFyFzDU2q9vGf4B8P3oRz+KyD33+Xy+X/7yl8p9t3Ol33uYhChEysR39uzZiDxwP/7xj5UH7l2hr5qB5YBv+PDhvr6+vojc96Dgp1yK4z2+c/gS8P30pz81/WFzu92+9PR05YGTE4b7bsKJUJ/xlZaWmn7fHz586HM4HD4EK1Kylc17yMLOhJlg7IHIqM4cPXqU5uZmgGa+38QLL/77HgnyzPbt2/H5fCC4E09DHP4e31EkA32A7969e6auLD/4wQ+U1fx/WvuVbYFVgG/kyJGmx0fmzZun3HeZlNp7fIdxEvD94he/MO1h6+zs9A0ZMkR54OTbfH53EY1Qr/WdPHnStPt+9+5d5Z67gXRLv/F3FO+K6w7+GnUzq9l2795Nd3c3wG1Eg73vO/rx33czS1eVnmrAKcSW6T2+x8gBfDExMb729nZTVpaSkhJlZflLi7+rnbAO8I0ePdrn8XhMue/Tp09X7vv/afF3fQ+bwDSWXENDQzBZY4LF39NOiAPaAV9FRYXh972qqkox8l5guKXf9DuMd8l1B3/Zp9Lk0Ej87ne/E22NoQKoM/wE7y568d93M4QjlUo1BNVYU7vT9/juwjSW3KxZs5SV5Y8s/o52xBbAl52d7fN6vYbe9yCq8f9h8Xf8TkNOWtM+iEIEa0b8y7/8i2EdV2/dusXmzZtBpPAyEa2K3yOABMR9T/r666+ZN2+eIZNevXqVTz75BERHlXSgy5CJ3+M7gd+iv1Y51L/vfdfNQbAN8+67fcTkv6Owez36QNiFiASbgX8xad7vArYhCDRm4CuT5n0PP/5/x3WwVy9a2hsAAAAASUVORK5CYII=" />
```
foolip: Is this just for fun, or can the image be omitted?

run/identify-sha.sh
```diff
@@ -0,0 +1,14 @@
+# Copyright 2018 The WPT Dashboard Project. All rights reserved.
+
+# Use of this source code is governed by a BSD-style license
+# that can be found in the LICENSE file.
+
+# Run this script in cron, and have it write to the
+# /var/www/html folder a file called sha
+#
+# You should have Options +Includes, mod_include and XBitHack on in
```
foolip: Can you make this part of some checked-in config?

run/identify-sha.sh
```diff
@@ -0,0 +1,14 @@
...
+# your <VirtualHost> for the htdocs/index.html page to work.
+
+sha="$(git ls-remote https://github.com/w3c/web-platform-tests.git | head -1 | cut -f 1)"
```
foolip: I guess this is a placeholder? I think that for external runs, right now looking for the latest run in https://wpt.fyi/test-runs with "many" browsers would be a better bet, but an API that just tells you what to try running and in which order would be nicer.

run/htdocs/index.html
mattl: Yeah, just for fun.

run/identify-sha.sh
mattl: Sure, I'll add this.

run/identify-sha.sh
mattl: This would be where runners would look to populate that test-runs page. 

Every browser could `curl https://sha.wpt.fyi/sha` and get a SHA and pass that to run.py or whatever they're using.

foolip: The challenge with external runs ("outside of the wpt.fyi infrastructure") is how to test the same commit of wpt as has already been tested on wpt.fyi, or is in the process of being tested. https://github.com/w3c/wptdashboard/issues/292 has some links to related issues.

If one is OK with always being behind, https://wpt.fyi/api/runs?max-count=100 is a decent API for this already. To improve upon that, I think that the commit to be tested needs to be published while it is being tested. Then one could either use that and https://wpt.fyi/api/runs?max-count=100 to determine which commits to test, or a single feed of commits to test could be provided. If it's a single feed, then possible some notion of priority is needed, so that external runs that runs just daily/weekly/monthly will tend to use the same commit.

jugglinmike:
We addressed this issue via the introduction of a Buildbot-driven infrastructure (see gh-522).

Issue #456, opened 2018-02-13
wpt.fyi site deployment is broken


rwaldron:
@lukebjerring @mdittmer 

```
Updating service [default]...failed.
ERROR: (gcloud.app.deploy) Error Response: [9] Deployment contains files that cannot be compiled: Compile failed:
/work_dir/anomaly_handler.go:34: undefined: metrics.GetDatastoreKindName
/work_dir/anomaly_handler.go:35: undefined: metrics.PassRateMetadata
/work_dir/anomaly_handler.go:37: undefined: metrics.PassRateMetadata
/work_dir/anomaly_handler.go:69: undefined: metrics.GetDatastoreKindName
/work_dir/anomaly_handler.go:70: undefined: metrics.FailuresMetadata
/work_dir/anomaly_handler.go:74: undefined: metrics.FailuresMetadata
/work_dir/interop_handler.go:20: undefined: metrics.GetDatastoreKindName
/work_dir/interop_handler.go:20: undefined: metrics.PassRateMetadata
/work_dir/interop_handler.go:23: undefined: metrics.PassRateMetadata
2018/02/12 16:50:28 go-app-builder: build timing: 16×compile (4.175s total), 0×link (0s total)
2018/02/12 16:50:28 go-app-builder: failed running compile: exit status 2
```

foolip:
@lukebjerring was telling me yesterday that something odd had happened to metrics data in prod, so this might be a consequence of that?

lukebjerring:
No; that was with the metrics computation code. This looks like an import name mismatch.

lukebjerring:
I just sync'd to head, and ran `./git/git-deploy` and it succeeded:
https://lukebjerring-upstream-master-dot-here-lift-this.appspot.com

It's likely that your local $GOPATH/src/github.com/w3c/wptdashboard folder is a symlink and there's something wrong with your imports. Have you tried deploying from the cloud console?

mariestaver:
@lukebjerring @rwaldron is deployment for wptdashboard fixed?

rwaldron:
@mariestaver I have an open PR to add installing the dependencies to the deployment process. @jugglinmike is reviewing.

rwaldron:
Issue run down: 

1. For local development, the go deps are fetched and installed within the docker instance
2. The meager tests written for anomaly and interop (in main_test.go) are run in the docker instance
3. The site is run from the docker instance for development. 
[...]

PR #468, opened 2018-02-15
Ensure webapp golang deps are installed before deploying. Fixes gh-456

rwaldron:


rwaldron:
```
./util/deploy.sh --version=excellent
```

Press "n" when prompted to proceed.


To check the deps:

```
cd $(go env GOPATH)/pkg/
```

(you'll have to navigate from there on your own)

util/deploy.sh
```diff
@@ -18,6 +18,11 @@ while getopts ':ph' flag; do
   esac
 done
 
+# Ensure dependencies are installed.
+info "Installing dependencies..."
+cd ${WPTD_PATH}; make go_deps;
```
mdittmer: Is there a chance that `cd` will have unwanted side effects for the rest of the script? Should we have `pushd ${WPTD_PATH} && make go_deps && popd`?

rwaldron: No, I don't think it will: the next set of tasks are git-centric information gathering and then finally `gcloud app deploy --no-promote --version=w3c-move-deploy-to-utils /absolute/path/to/wptdashboard/webapp`.

rwaldron:
@mdittmer thanks again for the review. Landed: https://github.com/w3c/wptdashboard/commit/2e1e4b09450724b12a39ab46529ac2f079b72baf

PR #467, opened 2018-02-15
run.py: Implement `--partial-threshold` argument

jugglinmike:
This is intended to resolve gh-465

PR #469, opened 2018-02-15
Implement workaround for bug in WPT CLI

jugglinmike:
@rwaldron This is a fix that I originally implemented as part of gh-467. I've decided to backport it to `master` for a few reasons:

- I have new evidence that the WPT CLI bug is currently preventing us from benefiting from the "retry" heuristic implemented in gh-426
- gh-467 is intended to implement an unrelated feature and includes a lot of change. It's review should not be hastened by our desire to fix the bug

So I recommend that we focus on this patch first. (Rebasing gh-467 will be trivial.)

### DIFF
--- a/run/run.py
+++ b/run/run.py
@@ -216,7 +216,15 @@ def main(platform_id, platform, args, config):
     full_report = {'results': []}
     for name in glob.glob('%s/*.json' % abs_report_chunks_path):
         with open(name) as f:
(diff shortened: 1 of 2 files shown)


rwaldron:
Rebased and manually merged 11d55825a6f8be0f167e84bfd30d29f9a6702272

PR #471, opened 2018-02-17
Load browser config. data during initialization

jugglinmike:
Previously, data describing the available browsers was not loaded until
the web server handled a request which required it. Subsequent requests
were intended to re-use the result of this operation, persisted as
global application state.

That strategy was susceptible to a race condition: if the initial
request that triggered data loading was quickly followed by a similar
request, the loading logic would be executed a second time, and the
resuting in-memory representation would include two entries for each
available browser.

Re-factor the internals to load the browser data at module
initialization time (e.g. before any requests are received).

### DIFF
--- a/webapp/util.go
+++ b/webapp/util.go
@@ -12,59 +12,39 @@ import (
 	models "github.com/w3c/wptdashboard/shared"
 )
 
-var browsers map[string]models.Browser
-
-// No 'set' type in Go, so use map instead.
-var browserNames map[string]bool
-var browserNamesAlphabetical []string
-
-// GetBrowsers loads, parses and returns the set of names of browsers
-// which are to be included (flagged as initially_loaded in the JSON).
-func GetBrowsers() (map[string]models.Browser, error) {
-	if browsers != nil {
-		return browsers, nil
-	}
-	var bytes []byte
-	var err error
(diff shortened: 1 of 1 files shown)


webapp/util.go
```diff
@@ -12,59 +12,39 @@ import (
...
-	if bytes, err = ioutil.ReadFile("browsers.json"); err != nil {
-		return nil, err
-	}
-
-	if err = json.Unmarshal(bytes, &browsers); err != nil {
-		return nil, err
-	}
-	return browsers, nil
-}
+var browserNames, browserNamesAlphabetical = loadBrowsers()
```
mdittmer: nit: Here or above `loadBrowsers()` implementation, call out that any error is considered fatal. @lukebjerring do you know what AppEngine will do if instances keep panicking on startup? I assume it gives up after a few retries and somehow flags the faulty deployed version, but I've never encountered this before.

rwaldron:
@jugglinmike I'm going to merge this and create another commit for adding the documentation that @mdittmer requested.

webapp/util.go
lukebjerring: Locally, it seems to try a (large, but) finite number of times and then fall back to serving a copy of the failing stacktrace as a 500 for all requests. I think a unit test to ensure it parses is sufficient to catch this issue before prod..

Issue #472, opened 2018-02-19
`run.py` fails to validate test results when using Sauce Labs


jugglinmike:
Results are not being automatically reported for the Edge and Safari browsers.  For those browsers, the tests are running to completion, but the process exits prior to data upload when the result set is being validated.

[I recently introduced that validation logic](https://github.com/w3c/wptdashboard/pull/467) under the erroneous assumption that `./wpt run --list-tests sauce` would function even though no browser is installed locally. This is not the case, and [I verified with James Graham that the requirement is intentional](https://w3.logbot.info/testing/20180219#c379710-c379712):

>     (11:18:05 AM) jgraham: jugglinmike: I think you need to pass in a product
>                            to make the code that loads the tests work
>     (11:19:24 AM) jgraham: The reason for *that* is that you can
>                            theoretically have product specific metadata that
>                            would alter the set of tests run

James has helped me design an alternate approach that should work for any browser configuration. I am working on patch presently.

PR #475, opened 2018-02-20
run.py: Re-implement results validation

jugglinmike:
Previously, `run.py` was extended to ensure that the actual result set
described the expected number of tests (within some threshold). This was
implemented via the `--list-tests` flag provided by the WPT CLI.

Unfortunately, this functionality requires that the web browser under
test be installed on the local system. This is not the case when running
tests remotely using WebDriver, so the new validation logic would fail
in those cases.

The same information can be retrieved from a previously-unused output
stream (the "raw log") provided by WPT during test running.

Re-implement results validation to instead inspect the structured output
of the test run itself.

This is intended to resolve gh-472.

### DIFF
--- a/run/run.py
+++ b/run/run.py
@@ -17,6 +17,7 @@ import re
 import requests
 import shas
 import subprocess
+import tempfile
 import traceback
 import os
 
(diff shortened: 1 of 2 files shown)


rwaldron:
👍

PR #480, opened 2018-02-20
Update tests

rwaldron:
To run locally: 

```
docker exec -it -u $(id -u $USER):$(id -g $USER) wptd-dev-instance python run/run_test.py
```

```
docker exec -it -u $(id -u $USER):$(id -g $USER) wptd-dev-instance python run/shas_test.py
```

### DIFF
--- a/run/shas_test.py
+++ b/run/shas_test.py
@@ -8,6 +8,7 @@ import logging
 import mock
 import os
 import shas
+import shutil
 import subprocess
 import unittest
 
@@ -15,6 +16,17 @@ from datetime import date
 
 here = os.path.dirname(__file__)
 wptd_dir = os.path.join(here, '../')
+target_dir = os.path.abspath(os.path.join(os.path.abspath(wptd_dir), '../', 'wptdashboard-temp'))
+
+
+if os.path.exists(target_dir):
+    print('%s exists, so it will be deleted and recloned for this test' % target_dir)
+    shutil.rmtree(target_dir)
+
+os.makedirs(target_dir)
+command = ['git', 'clone', '--depth', '1', 'https://github.com/w3c/wptdashboard', target_dir]
+return_code = subprocess.check_call(command, cwd=wptd_dir)
+assert return_code == 0, ('Got non-0 return code: %d from command %s' % (return_code, command))
 
 
 class TestSHAFinder(unittest.TestCase):
@@ -27,7 +39,7 @@ class TestSHAFinder(unittest.TestCase):
             '--unshallow',
         ]
         abspath = os.path.abspath(wptd_dir)
-        subprocess.call(command, cwd=abspath)
+        subprocess.call(command, cwd=target_dir)
 
         # ~5 commits that day, ensure first is result.
         logger = mock.Mock(logging.Logger)
--- a/run/shas_test.py
+++ b/run/shas_test.py
@@ -14,21 +14,28 @@ import unittest
 
 from datetime import date
 
-here = os.path.dirname(__file__)
-wptd_dir = os.path.join(here, '../')
-target_dir = os.path.abspath(os.path.join(os.path.abspath(wptd_dir), '../', 'wptdashboard-temp'))
-
-
-if os.path.exists(target_dir):
-    print('%s exists, so it will be deleted and recloned for this test' % target_dir)
-    shutil.rmtree(target_dir)
-
-command = ['git', 'clone', '--depth', '1', 'https://github.com/w3c/wptdashboard', target_dir]
-return_code = subprocess.check_call(command, cwd=wptd_dir)
-assert return_code == 0, ('Got non-0 return code: %d from command %s' % (return_code, command))
-
 
 class TestSHAFinder(unittest.TestCase):
+    here = os.path.dirname(__file__)
+    wptd_dir = os.path.join(here, '../')
+    target_dir = os.path.abspath(
+        os.path.join(os.path.abspath(wptd_dir), '../', 'wptdashboard-temp')
+    )
+    def setUp(self):
+        # This is only necessary for local development environments
+        # that are unlikely to have been cloned with an explicit --depth
+        if os.path.exists(self.target_dir):
+            shutil.rmtree(self.target_dir)
+
+        command = [
+            'git', 'clone', '--depth', '1', 'https://github.com/w3c/wptdashboard',
+            self.target_dir
+        ]
+        return_code = subprocess.check_call(command, cwd=self.wptd_dir)
+        assert return_code == 0, (
+            'Got non-0 return code: %d from command %s' % (return_code, command)
+        )
+
 
     def test_nov_21st(self):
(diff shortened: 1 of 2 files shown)


run/shas_test.py
```diff
@@ -8,13 +8,25 @@
...
+
+
+if os.path.exists(target_dir):
+    print('%s exists, so it will be deleted and recloned for this test' % target_dir)
+    shutil.rmtree(target_dir)
+
+os.makedirs(target_dir)
+command = ['git', 'clone', '--depth', '1', 'https://github.com/w3c/wptdashboard', target_dir]
+return_code = subprocess.check_call(command, cwd=wptd_dir)
+assert return_code == 0, ('Got non-0 return code: %d from command %s' % (return_code, command))
```
jugglinmike: The problem here is that TravisCI does a shallow clone by default, but that's not the case during local development. By replicating the issue in the development environment, this solution treats TravisCI like a lowest-common denominator. What do you think about fixing the problem in instead? [We could do that declaratively for all builds](https://docs.travis-ci.com/user/customizing-the-build/#Git-Clone-Depth) or if efficiency is a concern, we could do it during the "install" phase just for the Python unit test build.

That said, I think any test which modifies itself is asking for trouble. If the process ends abruptly, the container is going to be in a weird state. Your idea to create a copy of the repository specifically for testing purposes saves us from that. My recommendation would be to solve the "shallow repo" problem using one of the two methods above, and also avoiding state corruption by using the approach you've laid out here (just using the `shutil.copytree` method instead of `git clone`). This second part is beyond your original scope, though, so I'd be happy to do it myself in a follow-up patch.

jugglinmike:
Thanks for fixing this, @rwaldron! It looks like I introduced a regression in the Python test suite, but we weren't aware of that because we've been testing functionally (these failures do not reflect issues with the application's behavior), and because [TravisCI was failing silently](https://travis-ci.org/w3c/wptdashboard/jobs/341104830#L3690-L3703).

I've submitted gh-481 so we're made aware of these issues automatically (but we'll need to land your fix first).

run/shas_test.py
rwaldron: Thanks for the notes, just a few questions: 

> using the shutil.copytree method instead of git clone

This is a good strategy for working directory repositories that were cloned with some explicit `--depth`, as this is for TravisCI. The shallowness of that repo can be forever preserved by making a copy before doing the `git fetch --unshallow` is ever encountered in these tests. It's unlikely (and unknowable) that a development clone will be created with any explicit `--depth`, which means that making a copy will produce the same result: 

```
rwaldron in ~
$ git clone git@github.com:w3c/wptdashboard.git wptd && cp -r wptd/ wptd-copy && cd wptd-copy && git fetch --unshallow
Cloning into 'wptd'...
remote: Counting objects: 2447, done.
remote: Compressing objects: 100% (20/20), done.
remote: Total 2447 (delta 8), reused 10 (delta 6), pack-reused 2421
Receiving objects: 100% (2447/2447), 8.49 MiB | 2.83 MiB/s, done.
Resolving deltas: 100% (1227/1227), done.
fatal: --unshallow on a complete repository does not make sense
rwaldron in ~/wptd-copy on master
```

Here's the same operation with an explicit `--depth`: 

```
rwaldron in ~
$ git clone --depth=50 git@github.com:w3c/wptdashboard.git wptd && cp -r wptd/ wptd-copy && cd wptd-copy && git fetch --unshallow
Cloning into 'wptd'...
remote: Counting objects: 645, done.
remote: Compressing objects: 100% (386/386), done.
remote: Total 645 (delta 262), reused 550 (delta 229), pack-reused 0
Receiving objects: 100% (645/645), 5.84 MiB | 7.22 MiB/s, done.
Resolving deltas: 100% (262/262), done.
remote: Counting objects: 1187, done.
remote: Compressing objects: 100% (445/445), done.
remote: Total 1187 (delta 802), reused 1082 (delta 698), pack-reused 0
Receiving objects: 100% (1187/1187), 1.32 MiB | 1.34 MiB/s, done.
Resolving deltas: 100% (802/802), completed with 72 local objects.
rwaldron in ~/wptd-copy on master
```


All of that aside, I _do_ see your concern about potentially screwing up the container's state, but I need to think about that a bit more

rwaldron:
@jugglinmike w/r to my comment: 

> All of that aside, I do see your concern about potentially screwing up the container's state, but I need to think about that a bit more

What if we only do this for non-CI test runs: 

```python
# This is only necessary for local development environments
# that are unlikely to have been cloned with an explicit --depth
if (os.getenv("CI") == None):
    target_dir = os.path.abspath(
        os.path.join(os.path.abspath(wptd_dir), '../', 'wptdashboard-temp')
    )


    if os.path.exists(target_dir):
        print(
            '%s exists, so it will be deleted and recloned for this test'
            % target_dir
        )
        shutil.rmtree(target_dir)

    command = [
        'git', 'clone', '--depth', '1', 'https://github.com/w3c/wptdashboard',
        target_dir
    ]
    return_code = subprocess.check_call(command, cwd=wptd_dir)
    assert return_code == 0, (
        'Got non-0 return code: %d from command %s' % (return_code, command)
    )
```

rwaldron:
Just a quick note: I have some lint stuff that's fixed in a new commit

jugglinmike:
In my mind, the ideal approach to testing this code would involve a git
repository that was created explicitly for this purpose. Don't worry, though:
I'm not going to suggest we take that route. I'm mostly saying that there are
many ways to skin a cat, and we can worry about optimizing the solution when we
have the luxury of time. Your solution is a pragmatic compromise between what's
needed and what's most correct.

I would like to change two things, though:

1. Nix the conditional for CI. We can put it back in if we find the builds are
   taking too long, but that seems unlikely. I think it's better to have
   consistency in this situation
2. Move the code to the `setUp` method of that file's `TestCast` sub-class.
   That way, the `unittest` will report problems as test failures (instead of
   just crashing altogether)

rwaldron:
@jugglinmike 

- [x] Nix the conditional for CI. We can put it back in if we find the builds are
taking too long, but that seems unlikely. I think it's better to have
consistency in this situation
- [x] Move the code to the setUp method of that file's TestCast sub-class.
That way, the unittest will report problems as test failures (instead of
just crashing altogether)

rwaldron:
Sans whitespace: https://github.com/w3c/wptdashboard/pull/480/commits/f7e695b27a87f858a1a112c5978c9bf97170eddd?w=1

jugglinmike:
Rebased and merged to `master` via the following commits:

- 735deba run/sha_test.py: move repo cloning into setUp
- 0ff3ccb run/sha_test.py: make fresh a clone whenever this test is run, prevents --unshallow error
- 68b52f4 run/run_test.py: remove test_report_to_summary test, patch_wpt (and stub_patch_wpt), report_to_summary

Nice work, Rick!

rwaldron:
Thanks again, Mike!

PR #481, opened 2018-02-21
Configure TravisCI builds to report test failures

jugglinmike:
The TravisCI documentation reads [1]:

> When one of the build commands returns a non-zero exit code, the Travis CI build runs the subsequent commands as well, and accumulates the build result.

Express disparate test commands as distinct items in an array within the TravisCI "script" configuration, ensuring that any failure causes the build to be labeled as "failing" (as opposed to only recognizing the result of the final command).

[1] https://docs.travis-ci.com/user/customizing-the-build/

This depends on gh-480

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,10 @@ install: |
   docker exec -u 0:0 "${DOCKER_INSTANCE}" chown -R $(id -u $USER):$(id -g $USER) /home/jenkins
 
 script:
-  - docker exec -t -u $(id -u $USER):$(id -g $USER) "${DOCKER_INSTANCE}" make "${MAKE_TEST_TARGET}"
+  - |
+    if [ "${MAKE_TEST_TARGET}" != "" ]; then
+      docker exec -t -u $(id -u $USER):$(id -g $USER) "${DOCKER_INSTANCE}" make "${MAKE_TEST_TARGET}"
+    fi
   - |
     if [ "${WCT}" == "1" ]; then
       cd webapp && npm test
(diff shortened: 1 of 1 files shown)


jugglinmike:
@rwaldron Now that we've landed 480, I've manually re-started the TravisCI build for this patch

jugglinmike:
@rwaldron Rebased onto lastest `master`. We'll see if this build fares any better

.travis.yml
```diff
@@ -75,12 +75,12 @@ install: |
...
   docker exec -u 0:0 "${DOCKER_INSTANCE}" chown -R $(id -u $USER):$(id -g $USER) /home/jenkins
 
-script: |
-  docker exec -t -u $(id -u $USER):$(id -g $USER) "${DOCKER_INSTANCE}" make "${MAKE_TEST_TARGET}"
-
-  if [ "${WCT}" == "1" ]; then
-    cd webapp && npm test
-  fi
+script:
+  - docker exec -t -u $(id -u $USER):$(id -g $USER) "${DOCKER_INSTANCE}" make "${MAKE_TEST_TARGET}"
```
rwaldron: Splitting these up exposed a bug that I introduced when I grouped them all together: since there is no `MAKE_TEST_TARGET` for the first entry of `matrix.include` array, this command doesn't need to be run—its failure was previously silenced by the command that follows. 

To fix this:

```js
- |
  if [ "${MAKE_TEST_TARGET}" != "" ]; then
    docker exec -t -u $(id -u $USER):$(id -g $USER) "${DOCKER_INSTANCE}" make "${MAKE_TEST_TARGET}"
  fi
```

Issue #430, opened 2018-02-05
`run.py`: Improve application logging


jugglinmike:
Extend the logging facilities of the `run.py` script:

- Record the standard output and standard error streams of `wpt` invocations to dedicated files
- Record information on partial trials
- Extend all output with timestamps

Depends on gh-428

See also gh-429

Issue #78, opened 2017-08-11
Prefix date to all log lines


jeffcarp:
When inspecting logs, the lines are not as useful or can be misleading since they aren't paired with a timestamp.

PR #482, opened 2018-02-21
run.py: Improve logging

jugglinmike:
Commit messages:

run.py: Report on missing tests & remove temp data

With the introduction of the "chunked execution" behavior, `run.py`
began generating intermediate results files for each test run (which
were subsequently compiled to create the complete test report). This
allowed for some additional forensics because operators could later use
those files to compare the tests that actually ran in a given chunk with
the tests that were expected to run. Unfortunately, this also
significantly increased the amount of disk space required to store each
result set.

Expand `run.py` to output information on "missing tests" for each
"chunk" and to remove all intermediate data files once the results have
been successfully reported.

---

run.py: enhance logging

Use the Python built-in `logging` module to prefix all output with a
timestamp.

---

That second commit resolves gh-430. It also resolves gh-78.

### DIFF
--- a/run/run.py
+++ b/run/run.py
@@ -19,6 +19,7 @@ import shas
 import shutil
 import subprocess
 import tempfile
+import threading
 import traceback
 import os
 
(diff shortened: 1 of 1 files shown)


rwaldron:
Excellent!

PR #488, opened 2018-02-22
Always use the firefox binary in running.ini

mattl:
This also includes an update to `browsers.json` which adds Firefox 58.0

### DIFF
--- a/webapp/browsers.json
+++ b/webapp/browsers.json
@@ -1,15 +1,15 @@
 {
     "chrome-64.0-linux": {
-        "initially_loaded": false,
-        "currently_run": false,
+        "initially_loaded": true,
+        "currently_run": true,
         "browser_name": "chrome",
         "browser_version": "64.0",
         "os_name": "linux",
         "os_version": "*"
     },
     "chrome-63.0-linux": {
-        "initially_loaded": true,
-        "currently_run": true,
+        "initially_loaded": false,
+        "currently_run": false,
         "browser_name": "chrome",
         "browser_version": "63.0",
         "os_name": "linux",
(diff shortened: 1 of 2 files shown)


rwaldron:
FWIW, The failures on TravisCI are not related to the changes in this PR

webapp/browsers.json
```diff
@@ -39,6 +39,14 @@
         "os_name": "linux",
         "os_version": "*"
     },
+    "firefox-58.0-linux": {
+        "initially_loaded": true,
+        "currently_run": true,
```
jugglinmike: I'm not sure how `currently_run` is used in the code (@rwaldron might be able to say more), but purely from a semantic standpoint, it seems like we should set the property to `false` for `firefox-57.0-linux`.

Similarly, this configuration of the `initially_loaded` property may cause results to be displayed for both versions of Firefox. It's less clear to me if that's desirable--on the one hand, we'd like to show the history data, but on the other hand, we probably don't need two stable releases of the same browser on the front page.

rwaldron: > it seems like we should set the property to false for firefox-57.0-linux.

Unless we plan to run them both, then yes, it should be. 

I forgot that this file is coupled tightly with different parts of the system. Thanks for taking a closer look at this.

mattl: Good call. I've updated the file accordingly. Let's see if this gives us a successful run before committing this.

mattl:
@jugglinmike I've updated this to support Firefox 59.0 and Chrome 64.

jugglinmike:
The test failures reported by TravisCI are unrelated to this changeset

jugglinmike:
Thanks, @mattl!

Issue #484, opened 2018-02-22
Fix lint issues currently causing the build to fail two jobs


rwaldron:

jugglinmike:
@rwaldron I think we can close this one

PR #497, opened 2018-02-26
run.py: satisfy linting requirements in test code

jugglinmike:


jugglinmike:
The build is failing due to previously-existing unit test failures. I will attempt to fix those tomorrow morning.

jugglinmike:
A number of unit tests were broken by gh-482, but we were not aware of this because of a misconfiguration in TravisCI [which has since been corrected](https://github.com/w3c/wptdashboard/pull/481). I've updated the tests. My solution is definitely not as flexible as it could be, but given the uncertain future of `run.py`, I don't think investing more time here is warranted.

@rwaldron: would you mind taking a look? I believe this resolves gh-484

rwaldron:
Thanks for resolving this @jugglinmike. LGTM

PR #506, opened 2018-03-02
Order the browsers alphabetically in browsers.json

foolip:
Chrome-Edge-Firefox-Safari, also matching the wpt.fyi UI.


foolip:
It's OK if this is left sitting until after March 9.

mariestaver:
Yup, we'll get to this after 3/9 -- Thanks @foolip !

rwaldron:
I've restarted the failed CI job

rwaldron:
Looks like that's all it needed!

jugglinmike:
This fell out-of-sync with `master`, where we have since inserted entries for the latest versions of Firefox (see gh-488). @foolip my fix is pretty uncontroversial, but I'm always skittish about modifying other people's commits. With your approval, I'll squash and merge this branch (note that the failure reported by TravisCI is unrelated and currently under investigation at gh-526).

foolip:
Please squash away :)

PR #515, opened 2018-03-13
Encrypted updated production file for four servers

mattl:
As mentioned, here is the encrypted file for the four servers.

Just running it on the Chrome machine right now, but tomorrow if things look good, I will try Firefox and then the Sauce machines.


mattl:
@jugglinmike Fixed the Safari version, bumped Firefox to 59.0. Chrome is already fixed by the earlier Firefox-paths PR.

jugglinmike:
We addressed this issue via the introduction of a Buildbot-driven infrastructure (see gh-522).

PR #517, opened 2018-03-14
Add HSTS header to every request

### LINKED ISSUE #128: Redirect http://wpt.fyi/ to https://wpt.fyi/ and use HSTS
See https://github.com/whatwg/misc-server/blob/master/debian/marquee/nginx/conf/whatwg-headers.conf for an example header.

jeffcarp:
This adds an HSTS header to every request that expires after 1 year.
It also applies the rule to all subdomains (we should check if this is
feasible) and opts-in to the HSTS preload list.

I'm not sure what the deploy process is like these days  but let me know how I can be of help getting this to production.

Resolves #128

### DIFF
--- a/webapp/main.go
+++ b/webapp/main.go
@@ -12,31 +12,45 @@ import (
 var templates = template.Must(template.ParseGlob("templates/*.html"))
 
 func init() {
-	// Test run results, viewed by browser (default view)
-	// For run results diff view, 'before' and 'after' params can be given.
-	http.HandleFunc("/", testHandler)
+	routes := map[string]http.HandlerFunc {
+		// Test run results, viewed by browser (default view)
+		// For run results diff view, 'before' and 'after' params can be given.
+		"/": testHandler,
 
-	// About wpt.fyi
-	http.HandleFunc("/about", aboutHandler)
+		// About wpt.fyi
+		"/about": aboutHandler,
 
-	// Test run results, viewed by pass-rate across the browsers
-	http.HandleFunc("/interop/", interopHandler)
+		// Test run results, viewed by pass-rate across the browsers
+		"/interop/": interopHandler,
 
-	// Lists of test run results which have poor interoperability
-	http.HandleFunc("/interop/anomalies", anomalyHandler)
+		// Lists of test run results which have poor interoperability
+		"/interop/anomalies": anomalyHandler,
 
-	// List of all test runs, by SHA[0:10]
-	http.HandleFunc("/test-runs", testRunsHandler)
+		// List of all test runs, by SHA[0:10]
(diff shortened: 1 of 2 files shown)


jeffcarp:
Added a couple unit tests

rwaldron:
Needs lint fix: https://travis-ci.org/w3c/wptdashboard/jobs/355592643#L3279

jugglinmike:
Hey @jeffcarp! I found a small typo in the new assertion function. This pull request has been lingering for a bit, so I pushed up a correction directly. I hope that's okay.

The only other issue is that we are now running http://builds.wpt.fyi, and that domain does not currently support TLS. I expect I can get that up and running within the span of a day, but I'd like to check in with @foolip before committing to that.

jeffcarp:
@jugglinmike thanks so much!

I was leaving this PR here since I didn't want to merge it and break anything (I should have been more explicit about that). Lmk if I can be of any further help.

foolip:
Is the cert for *.wpt.fyi all managed in a single place, or does build.wpt.fyi manage its own? Let's Encrypt with Apache plugin? That setup seems to work out of the box, whereas nginx needs an extra step to reload the server that is easy to forget even if the cert renewal is working.

Either way enabling HSTS seems fine.

jeffcarp:
@foolip wpt.fyi's certificate is [managed & auto-renewed by GAE](https://cloud.google.com/appengine/docs/standard/python/securing-custom-domains-with-ssl). That page says wildcard certs aren't supported right now.

jugglinmike:
build.wpt.fyi does not currently support HTTPS. Our options moving forward:

- Re-locate builds.wpt.fyi to some other domain
- Implement HTTPS support using Let's Encrypt
- Remove the `includeSubdomains` directive from the HSTS header introduced by this patch

The first alternative would be appropriate if there was some urgency in landing this patch. I don't get the sense this is the case. My preference is to go with the second since it's in our interest to secure the "build status" site in any event. I really only listed the third for completeness's sake :)

I can get HTTPS enabled for builds.wpt.fyi in the span of a day. This seems to be lower-priority than collecting results for unstable Chrome an Firefox, so I would plan to have it done by the end of next week (April 13). Does that sound good to you two?

jeffcarp:
Non-authoritative opinion: it sounds fine to ship without `includeSubdomains` now and then apply that in a future patch once builds.wpt.fyi is set up with TLS.

foolip:
@jeffcarp or @lukebjerring , can you move this PR over to https://github.com/web-platform-tests/wpt.fyi if it's still relevant?

jeffcarp:
Done https://github.com/web-platform-tests/wpt.fyi/pull/77

PR #526, opened 2018-03-27
Disable Chrome from WebComponentTester 

rwaldron:


jugglinmike:
Disabling the tests in Chrome seems a lot more drastic than pinning the version of an npm module. What change necessitated this if not a regression in a dependency?

jugglinmike:
(I ask because this pull request was initially submitted to modify the `package.json` file)

rwaldron:
> Disabling the tests in Chrome seems a lot more drastic than pinning the version of an npm module. 

Sorry, I forgot to update the description and title of this PR

> What change necessitated this if not a regression in a dependency?

I determined that this issue exists even when locking in the dependency version—something changed on the other end of the service we're relying on; I didn't have the time to deep dive into resolving the problematic remote configuration.

PR #529, opened 2018-04-02
Extend cleanup

jugglinmike:
It appears as though Firefox creates additional temporary files that we did not
previously account for. My guess is that these are only left behind in rare
circumstances because it has taken 13 days of continual test execution for a
significant number to accumulate.

This patch extends the existing solution for this problem by adding new entries
to a list of file name patterns which are deleted following every test execution.

This solution is proving not to generalize very well and is susceptible to new
problems with every browser release. In the future, I would like to investigate
alternatives. A heavy-handed approach would be to schedule a full system restart
following every build. Better still would be to execute builds from within LXC
containers which could be efficiently discarded.


jugglinmike:
Thanks, @rwaldron! These commits are available in [the `standalone-runner` branch](https://github.com/w3c/wptdashboard/tree/standalone-runner). I'll leave this open until we merge that to `master`.

jugglinmike:
The commits in this patch were merged via gh-538 back in April:

- f4de93a Remove additional temporary files
- bf25a1b Improve build step names

Issue #388, opened 2018-01-22
Nightly builds? Chrome or Chromium?


mattl:
Wondering what would be the best source of nightly builds for GNU/Linux, 64-bit? 

- Chromium has nightly builds, but my understanding is that they may have small differences like missing codecs, EME, etc?

- Chrome Canary doesn't exist for this platform, but there is a 'dev' channel.

foolip:
I think we should use Chrome Dev, since that's what we're using on Travis now.

To test vanilla Chromium builds built from the same commits as Chrome would be interesting for comparison with Chrome and other Chromium-based browsers, but I wouldn't say this is a priority, directly comparing results of (say) Chrome and QQ on Windows would mostly do the job.

mattl:
Good idea. @mariestaver what's a good place to make a note of this as a potential future low-effort browser to add to the stack?

mariestaver:
Snagged it for the backlog and applied appropriate labels -- we'll pull this work into a future sprint where adding a nightly browser makes sense (hopefully soon) -- Thanks for working this out, both of you!

mariestaver:
Also @mattl, I'm going to change the title to be more descriptive of the answer that came out of this discussion; feel free to edit further if needed.

mattl:
I'd like to start testing this with the new Ansible servers by not installing a stable Chrome on the machines.

mariestaver:
Moving to backlog as temporarily blocked by parallelization (to avoid creating unnecessary VMs) -- this may change after some further research/discussion; @mattl or @jugglinmike will update if so.  Thanks!

jugglinmike:
We are collecting results for Chrome's "dev" channel as of commit c8952d37985b44ebba85585f5436051cec425f4a . Since this project's focus was recently narrowed to only concern results collection, that commit resolves this issue.  The wpt.fyi website does not currently display this data, but that is no longer a consideration here. See https://github.com/web-platform-tests/wpt.fyi/issues/60 for details on the status of the feature on that website.

Issue #521, opened 2018-03-19
Add Firefox Nightly builds


mariestaver:

jugglinmike:
We are collecting results for Firefox Nightly as of commit c8952d37985b44ebba85585f5436051cec425f4a . Since this project's focus was recently narrowed to only concern results collection, that commit resolves this issue.  The wpt.fyi website does not currently display this data, but that is no longer a consideration here. See https://github.com/web-platform-tests/wpt.fyi/issues/60 for details on the status of the feature on that website.

PR #536, opened 2018-04-10
Install browsers dynamically&add exprmntl channels 

jugglinmike:
This is intended to resolve https://github.com/w3c/wptdashboard/issues/388 & https://github.com/w3c/wptdashboard/issues/521

### DIFF
--- a/src/scripts/get-browser-url.py
+++ b/src/scripts/get-browser-url.py
@@ -17,7 +17,7 @@ import urlparse
 import urllib
 
 
-def main(product, channel, os_name, bucket_name):
+def main(browser_name, channel, os_name, bucket_name):
     '''Find the most recent build of a given browser and provide a stable URL
     from which it may be downloaded. Because browser vendors do not necessarily
     commit to hosting outdated builds, this may involve downloading the build
@@ -29,11 +29,13 @@ def main(product, channel, os_name, bucket_name):
 
     source_url = None
 
-    logger.info('Locating artifact for %s browser', product.title())
+    logger.info(
+      'Locating artifact for %s@%s browser', browser_name.title(), channel
+    )
 
-    if product == 'firefox':
+    if browser_name == 'firefox':
         source_url = locate_firefox(channel)
-    elif product == 'chrome':
+    elif browser_name == 'chrome':
         source_url = locate_chrome(channel)
 
     if source_url is None:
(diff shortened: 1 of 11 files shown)


src/scripts/get-browser-url.py
```diff
@@ -0,0 +1,151 @@
...
+        subprocess.check_call([
+            'gsutil', 'cp', metadata_file, 'gs://%s.json' % uri
+        ])
+    finally:
+        os.remove(artifact_file)
+        os.remove(metadata_file)
+
+
+parser = argparse.ArgumentParser(description=main.__doc__)
+parser.add_argument('--product',
```
rwaldron: `--product` => `--browser-name`

test/upload_wpt_results_test.py
```diff
@@ -358,6 +367,7 @@ def test_duplicated_results(self):
             duplicated_results['1_of_2.json']['results'][0]
         )
         returncode, stdout, stderr = self.upload('firefox',
+                                                 'stable',
                                                  '1.0.1',
                                                  'linux',
                                                  '4.0',
```
rwaldron: Needs tests for `'experimental'`

src/scripts/upload-wpt-results.py
```diff
@@ -239,6 +250,7 @@ def notify(url, secret, payload):
 parser.add_argument('--raw-results-directory', required=True)
 parser.add_argument('--platform-id', required=True)
 parser.add_argument('--browser-name', required=True)
+parser.add_argument('--browser-channel', required=True)
```
rwaldron: Can you implement `choices: (...)` here as well? If so, then I recommend 👍

src/scripts/get-browser-url.py
jugglinmike: Definitely

src/scripts/upload-wpt-results.py
jugglinmike: You bet

test/upload_wpt_results_test.py
jugglinmike: Good catch :)

jugglinmike:
Squashed an merged to `standalone-runner` via the following commits:

- 2de45a0 Parameterize total chunks based on speed of build
- c8952d3 Install browsers dynamically&add exprmntl channels

Thanks, @rwaldron!

PR #537, opened 2018-04-11
Formally manage Google Cloud Platform inventory

jugglinmike:
Extend the project's Terraform configuration with descriptions of the
required object stores as provided by Google Cloud Platform.

@lukebjerring @hexcles I wanted to give you a heads-up about this. As long as
the results collector is uploading directly to the "wpt" bucket and providing
public URLs to wpt.fyi, I think it makes sense to manage its configuration
here. We won't be in this state for long, though. Once the runner is reporting
to a collector (as per Robert's diagramming on Monday), the "wpt" bucket will
become the collector's responsibility, and we can remove its entry in this
project's configuration.

In the mean time, if any changes are necessary, please make them here so that
I don't accidentally revert them while deploying.

### DIFF
--- a/provisioning/infrastructure/.gitignore
+++ b/provisioning/infrastructure/.gitignore
@@ -1,2 +1,3 @@
 *.plan
 /.terraform/
+/google-cloud-platform.json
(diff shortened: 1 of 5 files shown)


jugglinmike:
Merged to the `standalone-runner` branch at commit 43526f5a57e87b9af10153f60c05ae451508b879 .

Thanks, @lukebjerring!

PR #538, opened 2018-04-13
Remove relocated projects

jugglinmike:
After deciding [to split the codebases for results collection, data analysis, and web presence into three projects](https://github.com/web-platform-tests/results-collection/issues/533), we determined that the results collection codebase was the best candidate to inherit the existing project. This commit promotes that codebase to be and removes references to the other projects.

I performed this migration work shortly after the initial decision to split the projects. This branch includes the enhancements introduced and deployed since that time.

We're still in the process of migrating those other code bases, so it may be too early to merge this change set to `master`. @lukebjerring I'm comfortable continuing to work from this branch until things are ready on your end. I mostly wanted to give folks some more visibility into that work.


jugglinmike:
Thanks, @foolip!

PR #541, opened 2018-04-17
Report on any incomplete result sets

jugglinmike:
Previously, datasets which omitted results for some tests were
considered acceptable and subsequently uploaded to the web. The
underlying issues which motivated that design have been resolved, so the
project may be in a position to reject any incomplete datasets.

However, due to the imprecise nature of the previous validation
strategy, the actual completeness levels for certain browser
configurations are not well understood. Specifically, reports for
browsers which are mediated by the Sauce Labs service may continue to be
incomplete.

Decrease the threshold for missing tests to zero percent during results
validation. Maintain the prior threshold of two percent during results
uploading.

This configuration will surface any persisting issues with
incompleteness in the build system's web interface without interfering
with the overall reporting behavior. In this way, project maintainers
can verify that missing results are now a truly anomalous case, and a
subsequent commit can reject such datasets outright.

### DIFF
--- a/src/scripts/validate-wpt-results.py
+++ b/src/scripts/validate-wpt-results.py
@@ -8,8 +8,6 @@ import argparse
 import json
 import logging
 
-THRESHOLD = 0.02
-
 
 def main(log_wptreport, log_raw):
     '''Verify that the test results for a trial of the Web Platform Tests (as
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thank you, Philip!

PR #561, opened 2018-05-14
Update to latest version of Chromedriver

jugglinmike:
@kereliuk  We've fallen a few versions behind in Chromedriver. Nothing in the change log suggests there is any reason to wait, but there are a few fixes that might improve accuracy. Do you think this is good to go?


jugglinmike:
If you can remember, then you'd certainly be faster than me.

Alternatively, you could save yourself the effort (and solve this for other users) by publishing an RSS feed. It looks like the Chromedriver page is built on Google Sites; does that support RSS?

By the way, I noticed the supported Chrome versions are now being published with every release. Thanks for that!

Issue #565, opened 2018-05-29
Upload WPT report to staging.wpt.fyi


lukebjerring:
In preparation of migrating to the full-report based system on wpt.fyi, let's double-upload, sending the full report to staging.wpt.fyi for integration testing.

lukebjerring:
See https://github.com/web-platform-tests/wpt.fyi/issues/192 for wpt.fyi side of things.

Hexcles:
@jugglinmike I have a question for you:

In the [upload script](https://github.com/web-platform-tests/results-collection/blob/master/src/scripts/upload-wpt-results.py), do you already have the merged full report JSOn available at some point?

jugglinmike:
> In the [upload script](https://github.com/web-platform-tests/results-collection/blob/master/src/scripts/upload-wpt-results.py), do you already have the merged full report JSON available at some point?

Not currently. The "WPT report"-formatted results are stored on disk across many files, and they are individually parsed and stored in the legacy wpt.fyi format.

I avoided merging those files in the "WPT report" format because this would unnecessarily increase the memory footprint of the process. It's necessary now, though, so you can expect a patch to change this.

Here's what I'll need:

- the URL to which the data should be sent
- the expected HTTP method
- the expected HTTP headers
- the expected request body

Hexcles:
A code snippet is probably more helpful:

```python
# For Chrome & Firefox, report.json should have all the metadata.
requests.post(
    'https://wptdashboard-staging.appspot.com/api/results/upload',
    auth=('buildbot', 'secret_token'),
    files={'result_file': open('full_report.json.gz', 'rb')},
# For Sauce-based browsers you'd also need the following:
    data={'browser_name': ..., 'browser_version': ..., 'os_name': ..., 'os_version': ...}
)
```

Ideally I'd like to include the browser & OS info in the report even for Sauce. I'm looking into that.

jugglinmike:
Got it. Can you send me the token via e-mail? I will include it in the encrypted configuration file that's tracked within this repo

Hexcles:
@jugglinmike can you give me your pubkey?

jugglinmike:
Sure: https://keybase.io/jugglinmike

PR #566, opened 2018-05-31
Integrate with new results receiver 

jugglinmike:
@rwaldron @hexcles GitHub doesn't recognize the renames here, but I did this in two commits, so the change set should be easier to digest by viewing those individually.

The new CLI and code uses the property names from the WPT CLI's report. This is a little inconsistent with the "legacy" upload script, but I figured it would be more consistent in the long run.

This is intended to resolve gh-565


src/scripts/upload-wpt-results.py
```diff
@@ -47,205 +43,85 @@ def main(raw_results_directory, platform_id, browser_name, browser_channel,
...
-                             params={'secret': secret},
-                             data=json.dumps(payload))
+    # When the report is missing critical metadata, extend it with information
+    # provided via the command-line. (Currently, the WPT CLI does not include
+    # this metadata in reports generated via the Sauce Labs service.)
+    assert 'run_info' in report
+    assert isinstance(data['run_info'], object)
+    for name in (u'product', 'browser_version', 'os', 'os_version'):
+        if name not in report['run_info']:
+            report['run_info'][name] = platform_info[name]
```
Hexcles: This might not be correct.

I suspect Sauce runs do have 'product' (or maybe even other fields), but the wrong values (e.g. product=sauce, os=[client os]). Can you get a sample wptreport from a Sauce run? (Doesn't need to be the consolidated version; each shard should also have `run_info`.)

jugglinmike: You're right. The metadata is present and incorrect. Here's the report generated for a test run in Apple Safari via Sauce Labs:

```json
{
    "results": [
        {
            "message": "/infrastructure/assumptions/ahem.html d4a5ec8885c67c8491d08fa7b04268aba6a19480\n/infrastructure/assumptions/ahem-ref.html ba37c5363f2183149d7267dec4446634846b4721\nTesting d4a5ec8885c67c8491d08fa7b04268aba6a19480 == ba37c5363f2183149d7267dec4446634846b4721",
            "status": "FAIL",
            "subtests": [],
            "test": "/infrastructure/assumptions/ahem.html"
        }
    ],
    "run_info": {
        "bits": 64,
        "debug": false,
        "has_sandbox": true,
        "linux_distro": "Ubuntu",
        "os": "linux",
        "os_version": "16.04",
        "processor": "x86_64",
        "product": "sauce",
        "revision": "8c8238ca96f913c1b90f838def7f9cf0253c0113",
        "version": "Ubuntu 16.04"
    },
    "time_end": 1527784288550,
    "time_start": 1527784271236
}
```

[You've already offered to change this in WPT](), but I'd be glad to do it as
well.

Know that the build process is already aware of whether this report comes from
a Sauce-Labs-mediated browser. That means we have the information necessary to
address this without an upstream change. We should improve the WPT CLI
regardless, but that might take some time. If there's urgency here, I could
get this working by the end of the day.

Let me know how you'd like to proceed

jugglinmike: Forgot the link for "You've already offered to change this in WPT": https://github.com/web-platform-tests/wpt.fyi/pull/208#discussion_r191932905

Hexcles: @jugglinmike let's override the metadata for all Sauce runs and Sauce runs only (instead of filling in missing fields, which would also hide issues in Firefox/Chrome metadata that are supposed to be correct).

jugglinmike:
@Hexcles The latest commit introduces an option named `--override-platform` which explicitly causes the platform information to be inserted into the generated report.

src/scripts/upload-wpt-results.py
```diff
@@ -106,9 +112,9 @@ def consolidate(raw_results_files, platform_info):
     # this metadata in reports generated via the Sauce Labs service.)
     assert 'run_info' in report
     assert isinstance(data['run_info'], object)
-    for name in (u'product', 'browser_version', 'os', 'os_version'):
-        if name not in report['run_info']:
-            report['run_info'][name] = platform_info[name]
+    if platform_override is not None:
+        for name in platform_override:
+            report['run_info'][name] = platform_override[name]
```
Hexcles: Nit: you could drop `is not None`. Empty dict/lists are safe to be skipped, too.

And perhaps simply `report['run_info'].update(platform_override)`

jugglinmike: Much better!

jugglinmike:
Thanks for the review, @Hexcles! It's good to get a reality check, even if it's only cursory. And don't sell yourself short--you found a critical mistake.

Anyway, I had hoped to deploy this today, but due to [an earlier interruption in the collection process](https://github.com/web-platform-tests/results-collection/issues/568), we're a bit behind schedule for Edge results. Deploying this change will invalidate that collection, so I'd rather defer. Instead of deploying new code on a Friday, I'd like to target Monday morning--will that work for you?

Hexcles:
Thanks! Sure, Monday SGTM. Deploying on Friday is the recipe to kill an otherwise great weekend :)

jugglinmike:
Squashed and merged to `master` at the following commits:

- f5d090b Integrate with new results receiver
- 7e3c53f Reflect receiver's deprecated status in code

Issue #572, opened 2018-06-07
Send browser_channel as a label to results receiver


Hexcles:
From https://github.com/web-platform-tests/wpt.fyi/issues/245

```python
        response = requests.post(
            url,
            auth=(user_name, secret),
            data={'labels': browser_channel},
            files={'result_file': open(filename, 'rb')}
        )
```

Note: `labels` is a comma-separated string. Runners may add arbitrary labels and the wpt.fyi APIs can filter by labels.

cc @lukebjerring

lukebjerring:
Shouldn't labels be an array of strings? That would make it more consistent with the golang struct and  json marshalling behavior.

Hexcles:
@lukebjerring this is a HTML form not JSON, which doesn't really have arrays. It's more like the repeatable params we use in URLs.

PR #575, opened 2018-06-12
Include "labels" in payload sent to new receiver

jugglinmike:
This is intended to resolve gh-572. @Hexcles I believe that this extension is only relevant for the new results receiver. Is that right? Or should I amend this patch with an enhancement for the `wpt-legacy-wpt-results.py`?


jugglinmike:
No problem. Mind clicking the button for me?

PR #576, opened 2018-06-20
Switch over to results receiver in prod

Hexcles:
This commit atomically switches to results receiver to upload results to
prod and removes the legacy upload script.

We still keep uploading to staging.wpt.fyi for the ease of testing and
developing results receiver. Besides, the new results receiver is much
faster so there's minimal overhead.

## Review Information

@jugglinmike please take a *close* look at this PR. And I think we'd need a master restart to deploy this.

@foolip now that results receiver has been deployed in prod (https://github.com/web-platform-tests/wpt.fyi/issues/304), this is the **final** switch to launch the receiver.


jugglinmike:
Excellent, @Hexcles! This obviates a bunch of system configuration related to Google Cloud Platform, so I'm going to add a commit to remove that.

Based on your patch, it looks like the staging and production environments share the same "upload secret." Is that correct?

Hexcles:
@jugglinmike feel free to modify this PR branch. And yes, it's the same secret...

provisioning/configuration/roles/buildbot-master/tasks/main.yml
```diff
@@ -82,6 +82,8 @@
       value: '{{wptd_legacy_notify_secret}}'
     - name: wptd_upload_url
       value: '{{wptd_upload_url}}'
+    - name: wptd_staging_upload_url
+      value: '{{wptd_staging_upload_url}}'
```
jugglinmike: @Hexcles Buildbot has a notion of "secret" values which it will redact from publicly-visible logging output. It can read these values from flat files or from a Vault server. I chose the former to avoid the complexity that Vault would introduce, but it means that we need to maintain this unfortunate bit of duplication.

Hexcles:
@jugglinmike thanks for finishing this up! I'd appreciate it if you can ping me when you plan to deploy :)

jugglinmike:
@Hexcles and I have deployed this change and verified successful integration with both https://wpt.fyi/api/results/upload and https://staging.wpt.fyi/api/results/upload. :tada:

PR #577, opened 2018-06-21
Remove unnecessary property

jugglinmike:
The legacy results receiver required the date of the git revision under
test. Since this project no longer interfaces with that receiver, the
date no longer need to be tracked internally.


jugglinmike:
Thanks, .Robert!

PR #585, opened 2018-07-23
Update Mozilla Geckodriver

jugglinmike:
Version 0.21.0 of Mozilla Geckodriver was released on June 15, 2018 [1].
Update this project's provisioning scripts to install that version as
per [2].

[1] https://github.com/mozilla/geckodriver/releases/tag/v0.21.0
[2] https://github.com/web-platform-tests/wpt.fyi/issues/373

### DIFF
--- a/provisioning/configuration/roles/web-browsers/tasks/main.yml
+++ b/provisioning/configuration/roles/web-browsers/tasks/main.yml
@@ -47,7 +47,7 @@
 
 - name: Download Mozilla Geckodriver
   get_url:
-    url: https://github.com/mozilla/geckodriver/releases/download/v0.19.1/geckodriver-v0.19.1-linux64.tar.gz
+    url: https://github.com/mozilla/geckodriver/releases/download/v0.21.0/geckodriver-v0.21.0-linux64.tar.gz
     dest: ~/downloads/
 
 - name: Extend hosts file with configuration to shunt Firefox requests
--- a/provisioning/configuration/roles/web-browsers/tasks/main.yml
+++ b/provisioning/configuration/roles/web-browsers/tasks/main.yml
@@ -70,7 +70,7 @@
 
 - name: Install Mozilla Geckodriver
   unarchive:
-    url: ~/downloads/geckodriver-v0.21.0-linux64.tar.gz
+    src: ~/downloads/geckodriver-v0.21.0-linux64.tar.gz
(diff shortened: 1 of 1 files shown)


Hexcles:
I don't know why but I always had the impression that geckodriver (along with other dependencies) were automatically managed by `wpt` via `virtualenv`.

whimboo:
geckodriver is not a Python package but a compiled Rust executable. So it cannot be managed with virtualenv.

jgraham:
They are, but apparently the infrastructure overrides that.

Hexcles:
Ahh sorry my comment wasn't technically correct. I meant `wpt run` would download the latest `geckodriver` and `chromedriver` into `_venv/bin` (the virtualenv directory, but the drivers themselves have nothing to do with Python or virtualenv).

@jugglinmike why does the infra need to override the behaviour of `wpt run`?

jugglinmike:
@Hexcles We're manually installing the WebDriver binaries for largely historic
reasons. I'd be open to letting `wpt run` do it for us, but the patch will take
a little extra effort. I've opened gh-586 to explain the context of the current
approach, share some thoughts for the new approach, and track implementation
status. I'll merge this patch in the mean time since it has immediate value to
@whimboo.

jugglinmike:
Merged to `master` via commit f66de5515f828de75479474a59a5ba105c27a42d

whimboo:
Thanks a lot @jugglinmike!

whimboo:
Oh, how long will it take until it is active and used by runs as reported to wpt.fyi? The latest results from today still how those failures: https://wpt.fyi/results/webdriver/tests/get_element_property/get.py?sha=4ec9130408&label=experimental

I assume there is some deployment necessary first?

jugglinmike:
That's correct. I've now deployed the changes, so the next collection attempt (scheduled for 20:00 UTC) will use version 0.21.0 of Geckodriver.

In the interest of transparency, [I've also extended the deployment process to publish the active revision of this project](https://github.com/web-platform-tests/results-collection/commit/c6d961d19adda4b3dd13e8ef5a89b672af36b089). The page http://builds.wpt.fyi/#/about now displays the "results-collector version," which (today) is the git revision of the most recent deployment:

screenshot from 2018-07-24 13-51-39

It's not the most visible indication, I know, but it's something.

PR #593, opened 2018-08-27
Add a flag to the upload script to make timestamps optional

Hexcles:
## Description

The main purpose of this PR is to make timestamps (namely, `time_start` & `time_end` optional). These two timestamps are optional fields in the report (see [docs for results receiver](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md#results-creation)), and certain infra doesn't have these two fields in their reports, in which case they can use this flag to safely omit these two fields (cc @thejohnjansen ).

This is added behind a new flag (i.e. the timestamps are still required by default) because I don't want to change the default behaviour, and the two fields should be present if reports are produced by wptrunner.

## Changes

This PR also fixes some existing tests and removes a bunch of unused imports.


test/upload_wpt_results_test.py
```diff
@@ -366,7 +369,7 @@ def test_failed_request(self):
                                                  override_platform='false',
                                                  total_chunks=2)
 
-        self.assertNotEqual(returncode, 0, stdout)
+        self.assertNotEqual(returncode, 0, stderr)
```
jugglinmike: This test fails when the command exits successfully. In that case, standard error is less likely to have useful information.

src/scripts/upload-wpt-results.py
```diff
@@ -72,13 +68,16 @@ def main(raw_results_directory, product, browser_channel, browser_version,
                             'os_version': os_version
                         })
 
-            metadata['run_info'] = json.dumps(metadata['run_info'])
+            metadata_json = '"run_info":{}'.format(
```
jugglinmike: This value is not JSON, so I'd prefer to avoid that term. I can't think of a good alternative, so maybe instead we can just write to the handle directly.

Hexcles: Yeah it was a JSON in an earlier iteration. `metadata_bits`? `metadata_fields`?

Hexcles:
@jugglinmike I added a comment to say that the new flag, also the only optional flag, is for an external user. Besides, I tweaked the docstring to say that format has to meet the requirements of results receiver.

And regarding `--no-timestamps`, I think that's a bit too strict. "Optional" is more consistent with the requirement in the results receiver.

jugglinmike:
> And regarding `--no-timestamps`, I think that's a bit too strict.

I'm not familiar with the use case. Does the intended audience need to upload some files with timestamps and some files without timestamps?

> "Optional" is more consistent with the requirement in the results receiver.

That may be so, but it's not opt-in in the results receiver.

Hexcles:
@jugglinmike Hi Mike, I totally forgot this PR... I just rebased the PR and changed the flag from `--optional-timestamps` to `--no-timestamps` as you suggested. Can you take another look?

jugglinmike:
Squashed and merged to `master` at the following commits:

- e76f8b8 Print out stderr when tests fail
- 07a0cb0 Add a flag to make timestamps optional

Thanks, Robert!

Issue #591, opened 2018-08-15
Query "revision announcer" to select WPT revision


jugglinmike:
Currently, this project selects the revision of WPT to test according to the state of the HEAD reference of the git repository at the moment a collection attempt begins.

Recently, wpt.fyi was extended with a "revision announcer"--an HTTP endpoint which designates specific revisions of WPT for the purposes of coordination between multiple independent results collection systems.

Update this project to select the WPT revision by querying the revision announcer.

foolip:
https://wpt.fyi/api/revisions/latest is the API.

mdittmer:
If runners are locked to a certain cadence and need to "look back in time" to find the first not-yet-serviced commit, they can also use https://wpt.fyi/api/revisions/list

E.g.,

https://wpt.fyi/api/revisions/list?epochs=hourly&num_revisions=30

All epochs' metadata is at https://wpt.fyi/api/revisions/epochs

PR #596, opened 2018-08-28
Integrate with wpt.fyi "revision announcer"

jugglinmike:
## Description

Retrieve revisions of WPT in which to run tests from the wpt.fyi "revision announcer."

## Review Information

No change in operation.

## Changes

This should resolve gh-591.

## Requirements

This change introduces no new dependencies.

### DIFF
--- a/src/master/master.cfg
+++ b/src/master/master.cfg
@@ -124,17 +124,29 @@ c['schedulers'] = [
   schedulers.Nightly(name='Daily (remote builds)',
                      builderNames=['Chunk Initiator'],
                      onlyIfChanged=False,
-                     properties={'remote': True, 'worker_os': 'linux'},
+                     properties={
+                         'interval': 'daily',
+                         'remote': True,
+                         'worker_os': 'linux'
+                     },
                      hour=[12]),
   schedulers.Nightly(name='Four times per day (local GNU/Linux builds)',
                      builderNames=['Chunk Initiator'],
                      onlyIfChanged=False,
-                     properties={'remote': False, 'worker_os': 'linux'},
+                     properties={
+                         'interval': 'four_hourly',
+                         'remote': False,
+                         'worker_os': 'linux'
+                     },
                      hour=[0, 6, 12, 18]),
   schedulers.Nightly(name='Semi-daily (local macOS builds)',
                      builderNames=['Chunk Initiator'],
                      onlyIfChanged=False,
-                     properties={'remote': False, 'worker_os': 'macos'},
+                     properties={
+                         'interval': 'eight_hourly',
+                         'remote': False,
+                         'worker_os': 'macos'
+                     },
                      hour=[0, 12])
 ]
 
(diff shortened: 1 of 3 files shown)


src/master/master.cfg
```diff
@@ -205,18 +217,29 @@ for spec_id, spec in platform_manifest.iteritems():
...
+                                        sourceStamp={
+                                            # The `codebase` property must be
+                                            # set in order for this build step
+                                            # to function. It is currently
+                                            # undocumented:
+                                            # https://github.com/buildbot/buildbot/issues/4281
+                                            'codebase': u'',
+                                            'branch': None,
+                                            'repository': u'',
+                                            'revision': util.Property('announced_revision')
```
Hexcles: Is this the magic bit that makes BuildBot check out the correct revision?

src/scripts/get-wpt-revision.py
```diff
@@ -0,0 +1,52 @@
...
+
+import argparse
+import contextlib
+import httplib
+import json
+import urlparse
+
+
+@contextlib.contextmanager
+def request(method, url):
```
Hexcles: This seems a little complicated because of the low-level nature of `httplib`. Is `requests` not available here?

src/master/master.cfg
jugglinmike: It is. Took me over a day to figure it out :/

src/scripts/get-wpt-revision.py
jugglinmike: `requests` is not available here because I've been trying to limit external dependencies to Buildbot itself. That helps to make it possible to "drop in" these scripts, and it limits complexity in the project configuration (e.g. "install this `requirements.txt` when this script is present, install that `requirements.txt` when that script is present..."). The `request` function is called once and could easily be inlined; the separation is my attempt to encapsulate those low-level details so that the simplicity of the business logic is more obvious.

src/scripts/get-wpt-revision.py
```diff
@@ -0,0 +1,52 @@
...
+    with request('GET', 'https://wpt.fyi/api/revisions/latest') as response:
+        data = json.loads(response.read())
+
+    return data['revisions'][interval]['hash']
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description=main.__doc__)
+    parser.add_argument('interval',
+                        choices=('daily', 'eight_hourly', 'four_hourly',
```
mdittmer: nit: This looks like alpha order. Would duration-of-interval make more sense?

jugglinmike:
> Is there a plan to "catch up" if these builders go down? (E.g., temporarily over-provision to fill gaps.)

We that do in some cases, though the decision is orthogonal to integration with the revision announcer. We haven't formalized a policy due to a system limitation: we're not able to scale dynamically. A human operator has to decide when/if catching up is feasible.

> Is there a plan to ensure that longer intervals are hit in case these go down at a bad time?

Nope. That seems like a fine idea, but maintaining that state will require a fair amount of planning and new code.

jugglinmike:
Thanks, @Hexcles and @mdittmer!

jugglinmike:
[The latest results were collected for commit 93065b01a6](https://wpt.fyi/results/?sha=93065b01a6), as [selected by the revision announcer](https://wpt.fyi/api/revisions/list?at=2018-09-03T18:00:00Z), FYI ;)

Hexcles:
Great job!

mdittmer:
✋ 😂 ✋

PR #613, opened 2018-10-01
Collect results from stable Safari locally

jugglinmike:
This project was previously extended with support for collecting results
from Apple Safari Technology Preview via a worker running on a macOS
system. Extend the system to also support the stable release of Apple
Safari.

Previously, a single build step installed Safari Technology Preview and
enabled remote automation. Because the stable release of Safari is
bundled with the operating system, it cannot be installed. It does
require explicit enabling of remote automation, however. Refactor the
two operations into distinct build steps in order to support both
browsers.

Safari Technology Preview enables the "Develop" UI menu by default, so
the automation script could interact with it immediately. The menu is
not present in the default configuration of Safari. Add logic to ensure
the menu is available.

This project already included logic to determine whether automation had
successfully been enabled in Safari. That logic assumed the browser
implemented the W3C WebDriver protocol, but the stable release of Safari
does not. Extend the script to support both

This change is not supported by additional infrastructure. Decrease the
rate of collection in order to avoid over-utilizing available workers.

---

@foolip I know you're not too familiar with Buildbot internals, but I thought
you might be interested in seeing exactly how automation differs between Safari
stable and Safari Technology Preview.

I've vetted this change by deploying it to the results-collection staging
server last week. It's uploaded results from Safari's stable release a number
of times, though this is hard to recognize on wpt.fyi since we're also using
Sauce Labs to collect from Safari stable. It's easier see in [the most recent
upload](https://wpt.fyi/results/?sha=ee2e69bfb1&labels=stable) since I updated
to Safari 12 over the weekend (Sauce Labs has yet to support that release).

### DIFF
--- a/src/master/browsers.json
+++ b/src/master/browsers.json
@@ -47,6 +47,14 @@
         "os_version": "10.13",
         "remote": true
     },
+    "safari-11.1-macos-10.13": {
+        "browser_name": "safari",
+        "browser_channel": "stable",
+        "browser_version": null,
+        "os_name": "macos",
+        "os_version": "10.13",
+        "remote": false
+    },
     "safari-technology-preview": {
         "browser_name": "safari",
         "browser_channel": "experimental",
--- a/src/master/browsers.json
+++ b/src/master/browsers.json
@@ -47,7 +47,7 @@
         "os_version": "10.13",
         "remote": true
     },
-    "safari-11.1-macos-10.13": {
+    "safari-stable-macos": {
         "browser_name": "safari",
         "browser_channel": "stable",
         "browser_version": null,
--- a/src/master/master.cfg
+++ b/src/master/master.cfg
@@ -116,6 +116,11 @@ def render_chunked_builder(properties):
 
     return ['GNU/Linux Chunked Runner']
 
+@util.renderer
+def is_local_safari(properties):
+    return (properties.getProperty('browser_name') == 'safari' and
+            not properties.getProperty('use_sauce_labs'))
+
 c['schedulers'] = [
   schedulers.Triggerable(name='chunked',
                          builderNames=render_chunked_builder),
@@ -141,15 +146,15 @@ c['schedulers'] = [
                      },
                      hour=[0, 6, 12, 18],
                      minute=[5]),
-  schedulers.Nightly(name='Semi-daily (local macOS builds)',
+  schedulers.Nightly(name='Daily (local macOS builds)',
                      builderNames=['Chunk Initiator'],
                      onlyIfChanged=False,
                      properties={
-                         'interval': 'twelve_hourly',
+                         'interval': 'daily',
                          'remote': False,
                          'worker_os': 'macos'
                      },
-                     hour=[0, 12],
+                     hour=[0],
                      minute=[5])
 ]
 
@@ -172,8 +177,10 @@ for spec_id, spec in platform_manifest.iteritems():
     doStepIf = partial(filter_build, spec.get('os_name'), spec.get('remote'))
     browser_name = spec.get('browser_name')
     browser_channel = spec.get('browser_channel')
+    is_safari_stable = browser_name == 'safari' and browser_channel == 'stable'
+    should_install_browser = not spec.get('remote') and not is_safari_stable
 
-    if not spec.get('remote'):
+    if should_install_browser:
         name=str('Find installer for %s@%s' % (browser_name, browser_channel))
         locate_steps.append(
             steps.SetPropertyFromCommand(name=name,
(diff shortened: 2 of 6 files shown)


foolip:
Oh wow, we have Safari 12 results already!! \o/

Will review :)

src/master/browsers.json
```diff
@@ -47,6 +47,14 @@
         "os_version": "10.13",
         "remote": true
     },
+    "safari-11.1-macos-10.13": {
```
foolip: Is this really right? Isn't it Safari 12?

src/master/master.cfg
```diff
@@ -141,15 +146,15 @@ c['schedulers'] = [
                      },
                      hour=[0, 6, 12, 18],
                      minute=[5]),
-  schedulers.Nightly(name='Semi-daily (local macOS builds)',
+  schedulers.Nightly(name='Daily (local macOS builds)',
```
foolip: If the machine is fast enough, twice daily would be great :)

I was thinking about suggesting the Q4 goal for Safari stable+TP latency (from push to results) to be 24 hours, and that the average latency is 1.5 x cycle time.

src/scripts/safari-disable-popup-blocker.sh
```diff
@@ -0,0 +1,37 @@
...
+  done
+
+  echo Application defaults created. Terminating process.
+
+  kill -9 $safari_pid
+fi
+
+# source:
+# https://github.com/web-platform-tests/wpt/blob/1999770b55cb8cdd93dbce0e78e5c94b2ba22e0e/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
+defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true
```
foolip: Yeah, this is the sort of thing I had to do when trying to get `./wpt run safari` work :)

src/scripts/safari-enable-automation.sh
```diff
@@ -13,22 +13,46 @@
...
+if [ $1 == "stable" ]; then
+  browser_name="Safari"
+  safaridriver_binary="/usr/bin/safaridriver"
+else
+  browser_name="Safari Technology Preview"
+  safaridriver_binary="/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver"
+fi
 
+# Create a session using a body which satisfies both the Selenium JSON Wire
+# protocol (for Safari 11.1 and earlier) and the W3C WebDriver protocol (for
```
foolip: If we drop the Safari 11.1 support, is any simplification made possible? I guess that since you've already made it work, that's not really necessary though.

src/scripts/safari-enable-automation.sh
```diff
@@ -81,24 +108,45 @@ toggle_automation() {
               end if
           end repeat
 
+          -- Enable the "Develop" menu
+          click menu item "Preferences…" of menu applicationName of menu bar item applicationName of menu bar 1
```
foolip: This is a fascinating scripting language :)

src/master/browsers.json
jugglinmike: You're right. This technically works because the key doesn't influence any behavior. I'm going with `safari-stable-macos` to mirror the other entries and to limit the overhead of maintaining the correct version number since [I'm not very good at that](https://github.com/web-platform-tests/results-collection/commit/60a0bc91968ee4e3503508c7ea8f5d2aaa3db111#diff-e66b0e85335625519fa3847b00e226fd).

src/master/master.cfg
jugglinmike: > If the machine is fast enough, twice daily would be great :)

Unfortunately, the machine is not fast enough :/ More cause for more machines

> I was thinking about suggesting the Q4 goal for Safari stable+TP latency (from push to results) to be 24 hours, and that the average latency is 1.5 x cycle time.

This build runs once a day, but it includes both releases of Safari. That sounds like what you're describing here, though, so I think I'm misreading something. Would you mind re-phrasing?

src/scripts/safari-disable-popup-blocker.sh
jugglinmike: The trickiest part was recognizing that it does *not* work if you haven't run the browser before. That condition only occurs once for the life of the worker, but it's critical to producing a machine that just works.

src/scripts/safari-enable-automation.sh
jugglinmike: Yeah, the timing of these things may have obviated this workaround. Still, I'd like to keep it here. First: it allows us to revert to 11 if that's ever necessary. Second: it's become like a red badge of courage for me.

src/master/master.cfg
foolip: Starting a run once every 24 hours that takes 12 hours to complete means that the latency from PR being merge to results being available is in the 12-36 hour range. To get the worst-case latency to 24 hours it's necessary to run more often.

Although for simplicity we could pick a cadence (daily or 12h) and measure what percentage of those commits did get complete results.

jugglinmike:
Thanks, @foolip!

PR #618, opened 2018-10-15
Remove Safari 11.1

jugglinmike:
This project now collects results from Safari 12 via an
internally-managed macOS system. Results from that release are more
relevant to this project, so the overhead of maintaining integration
with the Safari 11.1 browser offered by Sauce Labs is no longer
justified.

Update the configuration so that the Buildbot build master no longer
creates jobs for Safari 11.1 via Sauce Labs.

### DIFF
--- a/src/master/browsers.json
+++ b/src/master/browsers.json
@@ -39,14 +39,6 @@
         "os_version": "*",
         "remote": false
     },
-    "safari-11.1-macos-10.13-sauce": {
-        "browser_name": "safari",
-        "browser_channel": "stable",
-        "browser_version": "11.1",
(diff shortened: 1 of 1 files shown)


jugglinmike:
:tada: :balloon: :pizza: :horse: :tada:

jugglinmike:
Would one of you folks mind merging on my behalf?

foolip:
Yep, seems like an appropriate amount of excitement for a browser update being completed :)

foolip:
@jugglinmike it seems like this wasn't effective, in https://wpt.fyi/test-runs we're still seeing Safari 11 results as recently as today.

PR #621, opened 2018-10-29
Turn Server-Timing experimental feature on

cvazac:
I'm attempting to improve the [server-timing results](https://wpt.fyi/results/server-timing) on Safari.

image

I will remove this if/when Safari enables this by default.

### DIFF
--- a/src/scripts/safari-disable-popup-blocker.sh
+++ b/src/scripts/safari-disable-popup-blocker.sh
@@ -31,6 +31,9 @@ fi
 # https://github.com/web-platform-tests/wpt/blob/1999770b55cb8cdd93dbce0e78e5c94b2ba22e0e/tools/wptrunner/wptrunner/browsers/sauce_setup/safari-prerun.sh
 defaults write com.apple.Safari com.apple.Safari.ContentPageGroupIdentifier.WebKit2JavaScriptCanOpenWindowsAutomatically -bool true
 
+# turn on experimental features
+defaults write com.apple.Safari ExperimentalServerTimingEnabled -bool true
+
 echo Closing all instances of the application to ensure the changes
 echo are observed.
 


cvazac:
@jugglinmike can you PTAL?

jugglinmike:
Thanks for the patch, @cvazac!

In order to help others reproduce the results they find on wpt.fyi, we should document this step more widely. Can you patch [WPT's instructions on running the tests in Safari](https://github.com/web-platform-tests/wpt/blob/master/docs/_running-tests/safari.md) (as published at https://web-platform-tests.org/running-tests/safari.html)?

As for the change itself, this may not the best place to insert the new setting. The name "safari-disable-popup-blocker.sh` doesn't describe a change like this. Also, note that this script only runs for the stable release of Safari. Should this configuration also apply to Safari Technology Preview?

I can help work this in to another script, though it'll be a few days before I have the time.

cvazac:
Thx for the review @jugglinmike!

I made a [new file](https://github.com/web-platform-tests/results-collection/pull/621/files#diff-62c40428ce5db7e89d43814770d43268) for the experimental flag code - but I couldn't figure out how to test it locally.

Yes, this should run in TP as well.

[PR](https://github.com/web-platform-tests/wpt/pull/13932)for safari.md documentation change.

src/scripts/safari-enable-experimental-features.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# turn on experimental features
+safari-enable-experimental-features.shsafari-enable-experimental-features.sh
```
Hexcles: Typo?

jugglinmike:
> Yes, this should run in TP as well.

Excuse me, I forgot that we have a precedent for this kind of thing.

[We only apply Chrome's `--enable-experimental-web-platform-features` when testing the browser's experimental release.](https://github.com/web-platform-tests/results-collection/blob/master/src/master/wpt_run_step.py#L119-L122) That makes the results for the "stable" release representative of a typical user's experience using the stock browser, and it also gives implementers visibility into the cutting edge. We should apply the same policy to Safari and Server Timing.

In addition to updating the script itself, there's another small change we'll need to make to support STP. We never know when the next release will be published, so in order to always test the latest version, we install the browser every time we attempt to collect results. In order for the new script to modify the freshly-installed browser, we'll have to invoke it after [that "build step" in the Buildbot configuration file](https://github.com/web-platform-tests/results-collection/blob/07f9839fb21a0f619403346664482a963757973e/src/master/master.cfg#L330-L336).

jugglinmike:
Also cc @foolip, who has been automating Safari in another environment.

foolip:
It would be great to get this into tools/wpt/run.py:
https://github.com/web-platform-tests/wpt/blob/bfc7ce508b754e84e95830c2186f1fa8eb9f179e/tools/wpt/run.py#L378-L385

See `Chrome.setup_kwargs` and `Firefox.setup_kwargs` in that file for precedent. If we put it into wpt itself, it'll also work for the Safari TP runs we'll be doing on PRs. cc @lukebjerring

foolip:
@cvazac @youennf are setting shared by Safari stable and TP, or are there settings that are only used by STP? Something that would make this easier is if it were possible to pass these settings to Safari as command line arguments to safaridriver. Do you think that'd make sense? cc @burg

burg:
They have separate settings. Safari is not big on adding command line arguments if possible. Would your use case be satisfied if experimental features could be toggled by WebDriver capabilities?

foolip:
Yes, that would be great I think, if they then have no  effect beyond the WebDriver session.

cvazac:
Picking back up on this... if I'm re-reading this correctly, we want to turn on some features for TP only. Has anything upstream made this easier - or should I make those changes from this PR?

cc @jugglinmike @foolip @burg

foolip:
@cvazac the code can live in wpt no problem, only issue is that it changes global state for a browser the user might also be using for browsing. Just putting it behind a flag that's enabled when running in CI would work for now.

foolip:
ping @cvazac, are you still interested in this?

cvazac:
@foolip Yes, I would love to clean up the Safari results. Unfortunately, the path to get there (^^^) is not at all clear to me. :(

foolip:
@cvazac is putting the behavior behind a flag that we only use in CI an OK path? Are there other blockers too?

foolip:
I've sent https://github.com/web-platform-tests/wpt/pull/16177 to do this for the Azure Pipelines setup which is now providing most of the results seen on wpt.fyi.

PR #633, opened 2018-11-26
Tolerate temporary unavailability of wpt.fyi

jugglinmike:
This project recently experienced a failure because wpt.fyi was
temporarily unavailable [1]. It has since been confirmed that
intermittent outages are expected [2], so the system should tolerate the
condition by retrying failed requests for a short period.

[1] https://github.com/web-platform-tests/results-collection/issues/632
[2] https://github.com/web-platform-tests/wpt.fyi/issues/802

### DIFF
--- a/src/scripts/get-wpt-revision.py
+++ b/src/scripts/get-wpt-revision.py
@@ -8,7 +8,12 @@ import argparse
 import contextlib
 import httplib
 import json
+import logging
+import time
 import urlparse
+
+
+logger = logging.getLogger(__name__)
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


jugglinmike:
While implementing this via Buildbot would make intermittent failures more apparent on [the web interface](http://builds.wpt.fyi), the framework currently [lacks support for a scripted "retry" behavior](https://github.com/buildbot/buildbot/issues/2864).

src/scripts/get-wpt-revision.py
```diff
@@ -32,17 +37,42 @@ def request(method, url):
...
+    if remaining <= 0:
+        logger.warn('Retry timeout exceeded.')
+        raise Exception(message)
+
+    logging.warn(message)
+    logging.warn('Retrying in %s seconds...' % (interval,))
+
+    time.sleep(interval)
+
+    with request_with_retry(method, url, interval, remaining) as response:
```
mdittmer: Is this magic naming that will delegate to `get_with_retry` on `method == 'GET'`? Is `method` a global? I don't see it anywhere else in the `get_with_retry` body.

mdittmer:
Added @Hexcles for review who is more familiar with this code base and Python in general.

src/scripts/get-wpt-revision.py
jugglinmike: This is a typo I introduced just before submitting the patch. Good catch!

src/scripts/get-wpt-revision.py
```diff
@@ -64,6 +94,18 @@ def main(interval):
...
     parser = argparse.ArgumentParser(description=main.__doc__)
+    parser.add_argument('--retry-interval',
+                        type=int,
+                        default=60,
+                        help='''Duration in seconds to wait between retrying
+                            failed requests to wpt.fyi'''
+                        )
+    parser.add_argument('--retry-timeout',
+                        type=int,
+                        default=60 * 1000,
```
Hexcles: Default timeout is, 1000 minutes? The timeout specified in`master.cfg` (600 seconds / 10 minutes) seems more reasonable. Or is this default value intentional?

jugglinmike: Another typo :(

jugglinmike:
Thanks!

PR #638, opened 2018-12-13
Upgrade pypi packages

Hexcles:
Mostly to update `requests` to fix a security vulnerability. Also update `pycodestyle` by the way.

### DIFF
--- a/src/scripts/get-binary-url.py
+++ b/src/scripts/get-binary-url.py
@@ -183,7 +183,7 @@ def locate_safari(channel):
         # selecting an incompatible binary intended for the Mojave release of
         # macOS
         match = re.search(
-            '(http[^\s]+\.dmg).*high\s*sierra', response.read(), re.IGNORECASE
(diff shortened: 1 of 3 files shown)


src/scripts/get-binary-url.py
```diff
@@ -183,7 +183,7 @@ def locate_safari(channel):
         # selecting an incompatible binary intended for the Mojave release of
         # macOS
         match = re.search(
-            '(http[^\s]+\.dmg).*high\s*sierra', response.read(), re.IGNORECASE
+            r'(http[^\s]+\.dmg).*high\s*sierra', response.read(), re.IGNORECASE
```
jugglinmike: Cool!

jugglinmike:
Requests' changelog is uncontroversial, so this looks good to me.

The Buildbot master also depends on that module at the outdated version. I'm guessing that GitHub.com isn't reporting it to us because the `requirements.txt` file is located in a subdirectory of this project. I'll update that in a follow-up commit.

Thanks, Robert!

PR #647, opened 2019-02-01
Switch from testing Edge 17 to 18

foolip:
Edge 44.17763.1.0 / EdgeHTML 18.17763 is available on Sauce Labs.

This is the latest stable version of Edge, and testing it will help validate
the setup for Edge on Azure Pipelines:
https://github.com/web-platform-tests/wpt/issues/14836

### DIFF
--- a/src/master/browsers.json
+++ b/src/master/browsers.json
@@ -15,10 +15,10 @@
         "os_version": "*",
         "remote": false
     },
-    "edge-17-windows-10-sauce": {
+    "edge-18-windows-10-sauce": {
         "browser_name": "edge",
         "browser_channel": "stable",
-        "browser_version": "17",
+        "browser_version": "18",
         "os_name": "windows",
         "os_version": "10",
         "remote": true


foolip:
@jugglinmike I'm not sure if there's anything that can be tested about this, but I presume that it'll require restarting the Buildbot master?

jugglinmike:
This looks good to me, but I'd like to wait on merging for a few days. We have [an open issue about an Edge-specific failure](https://github.com/web-platform-tests/results-collection/issues/644).  While my latest change seems to have fixed it, it's not conclusive. I'd like to see a few more successes before calling the problem resolved. With that, I'll be more comfortable making further changes to the configuration.

[It looks like Azure is configured to segment WPT into 20 "chunks,"](https://github.com/web-platform-tests/wpt/blob/92029dfc5158a888c836812a4cc50de8d030e6ec/.azure-pipelines.yml#L166) while [this project uses 100](https://github.com/web-platform-tests/results-collection/blob/5af91e8cc8026420f670025d3834c5182a5f98e3/src/master/master.cfg#L31).  For the purposes of comparison, [differences in test suite segmentation adds noise to the data](https://github.com/web-platform-tests/results-collection/issues/646#issuecomment-459427642).  Although my offer to change the Safari configuration stands, I'm less inclined to do the same for Edge. Even a segment as small as 1% of WPT requires a significant amount of time in Edge via Sauce Labs (as long as three hours).  With larger "chunks," intermittent failures will be much more costly, and we may not be able to recover.

foolip:
We could use more chunks in Azure Pipelines instead perhaps.

foolip:
@jugglinmike, if I do 100 chunks in Azure Pipelines, is this the setup I should replicate?
https://github.com/web-platform-tests/results-collection/blob/43658ddac1124f3028a10cbf0d137d78cc16e16c/src/master/wpt_run_step.py#L34-L46

That is, no special chunking?

foolip:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5841 is an attempted run with 100 jobs and the default chunk type. I expect it'll fail in some mysterious way, but we'll see.

jugglinmike:
@foolip see also the arguments we set for Sauce-Labs-mediated browsers:

https://github.com/web-platform-tests/results-collection/blob/43658ddac1124f3028a10cbf0d137d78cc16e16c/src/master/wpt_run_step.py#L71-L81

Most of those are irrelevant for Azure Pipelines, but that patch is missing `--run-by-dir`. I honestly don't know how that interacts with the "chunking" heuristic (it's a setting we inherited from [the original automation script](https://github.com/web-platform-tests/results-collection/blob/7a3febd9e82b009a32fc8205143e4508df9e24a7/run/run.py#L164)), so I'd just as soon remove it from this project.

Meanwhile, [that intermittent bug is still biting us](https://github.com/web-platform-tests/results-collection/issues/644#issuecomment-460341592), so I'm going to continue research using Edge 17.

jgraham:
`--run-by-dir` operates after chunking, so it doesn't make much sense if you're chunking by hash of full path. But it is intended to work well with the default directory-hash based chunking.

foolip:
Trying again with `--run-by-dir 3` in https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=5920

jugglinmike:
This necessitated restarting the Buildbot master, so I've done that.

